PR #2, opened 2016-01-27
My Nightly Update

ghost:
Remember that obscure quest rule Weapon Animation Fix we never got around to fully implementing. With this it actually does something! Also Diagonal Constant traps should be a thing, with angelscript simplification. I went ahead and did that.

### DIFF
--- a/output/common/angelscript/trapConstant.as
+++ b/output/common/angelscript/trapConstant.as
@@ -7,15 +7,82 @@ class TrapConstant: enemy
         mainguy=false;
         count_enemy=false;
         step=2;
+		dir=-1;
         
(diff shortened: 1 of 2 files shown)


Saffith:
Right now, the goal is not to break things. It's really not the time for new features.

Tamamo89:
Says the person who broke shops. :p

ghost:
Indeed, This is what happens when you release something too early. The code is such a mess right now, I wouldn't even commit any new features to this. Gleeok is right, this had to be done. Otherwise we would still be talking about it a year from now. As for fixing the spin attack. I'm going to be looking into that today.

Gleeok:
Make sure you compile and run any scripts first to make sure they actually compile!

PR #8, opened 2016-02-03
Current progress on cmake integration (DO NOT MERGE!)

lonegraywolf2000:
This is the start of a batch of commits meant to transition to using [CMake](http://cmake.org). At this time I have most of the Zelda Classic target set up in terms of source files. It does not compile due to allegro not knowing that VS2013/VS2015 includes modern headers.

The unity files have not been included yet. I haven't found the option to just include a source file and not compile it yet. There is some `HEADER_FILE_ONLY` flag, but I don't know if it works against `cpp` files.

Feel free to make suggestions or improvements in comments and forks as usual.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,14 @@
 *.elf
 *.ko
 
+# Generally ignored, but is meant as the main cmake directory.
+build/
+
+# Generated cMake files
+CMakeCache.txt
+CMakeFiles/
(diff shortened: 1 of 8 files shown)


evouga:
CMake is (finally!) going to be adopted. See the cmake branch for the current configuration-in-progress.

PR #12, opened 2016-02-05
Trying to allow VS2015 people to compile/link.

lonegraywolf2000:
Trying to get some stability so that I can make the right choices for #8.

At this time, it compiles, but not links.

Errors in general that are left (assuming debug mode compilation):
- Severity  Code    Description Project File    Line    Suppression State
  Error   LNK2038 mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in angelscript_unity.obj   zc  C:\Repos\zeldaWolfic\projects\vs2015\zc_as_module.obj   1  
- Severity  Code    Description Project File    Line    Suppression State
  Error   LNK2019 unresolved external symbol ___iob_func referenced in function _png_default_error    zc  C:\Repos\zeldaWolfic\projects\vs2015\libpng.lib(pngerror.obj)   1  
- Severity  Code    Description Project File    Line    Suppression State
  Error   LNK2001 unresolved external symbol ___iob_func  zc  C:\Repos\zeldaWolfic\projects\vs2015\libpng.lib(pngrutil.obj)   1  

Once these linking issues are resolved, we may be getting somewhere.


lonegraywolf2000:
...since I cannot assign anyone: @Gleeok, thoughts?

lonegraywolf2000:
@Gleeok Let's focus on here for now. The last issues are linker issues.

I changed the reference of `alld` to `alld_crt`. That resolved that one from Soundlib.

Soundlib also is missing the `_printf` symbol for some reason.

All projects are missing the `__iob_func` symbol.

Theoretically, once these missing symbols are found, I can start looking at cmake integration.

Gleeok:
Is this the same problem?
https://groups.google.com/forum/?_escaped_fragment_=topic/dislin-users/kagVl9LVFJA#!topic/dislin-users/kagVl9LVFJA

We usually compile without MSVCRT and link to LIBC if that helps. (/MT instead of /MTdll)

lonegraywolf2000:
@Gleeok That link hits it on the head from what I can tell.

The projects are already using `/MT` instead of `/MD`.

...welp. Kind of stumped. Wonder if we have to consider upgrading Allegro and its dependencies first. As far as compiling all of them...theoretically possible. Just a pain of gathering and bringing them in. Plus, that brings up a separate discussion of which dependencies should be kept in the repo and which should be submodules or some other option.

I'll need help on this one at this rate.

I'll make a separate issue request about Allegro 5 specifically since I did gather some research. I'll probably put it up tomorrow.

Gleeok:
Alright. I will try to help but get ready for a hack-attack!  (It is my specialty) :P

A) Does linking with  "legacy_stdio_definitions.lib" help?

B) This is actual serious proposal if the above does not work! :

[code] <code>
FILE _msvc15_sucks_balls[] = { _stdin, *stdout, *stderr };
extern "C" FILE_ __cdecl __iob_func(void) {
    return _msvc15_sucks_balls;
}
[/code] </code>

...I don't no how code tags work here...

edit: Or that might not work if they changed the FILE structure in 2015. .... Hmm....  Maybe only one dependency has to be updated unless there are more offending libs of course.

lonegraywolf2000:
A) Linking against the `legacy_studio_definitions.lib` file does not help. Still get the same linker errors.

B) Formatting uses three ticks. Adding a language also works after the first three: observe.

``` c++
#include <stdio.h>
FILE _msvc15_sucks_balls[] = { stdin, *stdout, *stderr };
extern "C" FILE __cdecl __iob_func(void) {
  return _msvc15_sucks_balls[0]; // Had to modify this line.
}
```

When the above code is placed in `sprite.cpp` as a sanity test, the zc project actually...well, compiled and linked. However, zquest returns duplicate linker errors. ...now that I think about it, I wonder if trying to compile the same source file against multiple projects could be an issue. See [StackOverflow](http://stackoverflow.com/a/15421368).

If I were to put the above code in a header file, say... `precompiled.h`, then the zc project gets the linker errors mentioned. In this case, the unity builds are working against us.

The proper fix would be to identify which files are needed across both ZC and ZQ, and move that code into its own zcommon library. This way, a proper Build Solution will work.

To go back to your question about the `FILE` structure...I don't know if it changed. It probably did since I modified your code to return the first value. If it matters, though...I think libpng is one of the primary factors. I suspect that recompiling that would. However, I've learned from experience that it's best to match the version of libpng to whatever Allegro intended. If this is to work, I need an exact version number of the libpng source to use so that it can be compiled in. zlib is probably similar: I want to say one depends on the other, but I don't recall which.

I'm not committing either A or B to this repo.

Gleeok:
If you only get the linker errors from libpng that might be the easiest thing to make it work. If not, maybe the above can be OK for now (as long as nothing crashes, anyway). 

Also zquest.cpp and zelda.cpp are definitely not shared for future reference.

header files _should_ contain the version information.

evouga:
Stale and obsoleted by CMake.

PR #14, opened 2016-02-07
get vs2015 building with all deps committed

zeromus:
Yes, it works. Yes, all deps are committed. You're wasting your life if you do it any other way. Trust me, I know what I'm doing.

PR #15, opened 2016-02-07
gitattributes not needed

lonegraywolf2000:


lonegraywolf2000:
@zeromus Got it covered for you. We can take further discussion on the remaining commits in #14 as before...or I guess I will go through the commit chaos and handle things there.

@Gleeok or @Saffith: Unless you know of a valid reason why we need `.gitattributes` anymore, we may need to consider merging this. We'll get the line endings back to a consistent spot...hopefully for good.

Gleeok:
From what we can gather .gitattributes is broken garbage anyway.

PR #22, opened 2016-06-23
Fixed ffcs jumping to changers

ghost:


Saffith:
That's not the right way to fix this. -1000 is a valid coordinate, so there's still a chance of it happening (albeit an extremely small chance). Better to ensure that prevX and prevY get set to -10000000 like they're expected to.

ghost:
Let's try again then!

Saffith:
Actually, can you even reproduce this in latest? I can't. Looks like it's been fixed already, so it's just 2.50 that still needs it.

Issue #4, opened 2016-01-27
Error trying compiling with vs 2008


vaualbus:
I can get the program compiled. I'm using vs2008 as the newer version give me even problem.
In the lib folder we miss alld.lib. If I download and use the allegro 4.2.2 version I get missed symbol.
__imp_pack_fopen_password() and cannot compile.
On zquest I get missing sysmbol from FFCSet::FFCSet, FFCSet:clear.

ghost:
Are you using the fixed version or your own version. You need to use the version AGN provides.

vaualbus:
ya but there no alld.lib in the libs folder. So I cannot use the modify
version. Where I can find the lib for the modified version?

2016-01-27 22:14 GMT+01:00 Meroune Lorelie notifications@github.com:

Gleeok:
Should be alld_crt.lib. Though I see there is no .dll for that version. I _think_ you don't even need the debug version. (Been a while since I had to debug allegro)

Will look into the zquest problem.

Gleeok:
VC++ builds should no longer require debug versions of allegro.

PR #24, opened 2016-07-12
Updated ffc swap (#4)

ghost:
- Updated ffc swap
- Updated zdefs.h
- Updated makefile
- Updated questRules.cpp

ffSWAPNEXT was bugged. I'm confident I fixed it. Also added a compatibility rule so the flag only effects speed. I haven't implemented the compatibility section into qst.cpp yet though.

### DIFF
--- a/makefile
+++ b/makefile
@@ -460,7 +460,7 @@ obj/enemyAttack.o: src/enemyAttack.cpp src/enemyAttack.h src/guys.h src/link.h s
 	$(CC) $(OPTS) $(CFLAG) -c src/enemyAttack.cpp -o obj/enemyAttack.o $(SFLAG) $(WINFLAG)
 obj/ffasm.o: src/ffasm.cpp src/ffasm.h src/ffscript.h src/gamedata.h src/jwin.h src/jwinfsel.h src/midi.h src/sprite.h src/tab_ctl.h src/zc_alleg.h src/zcmusic.h src/zdefs.h src/zquest.h src/zsys.h
 	$(CC) $(OPTS) $(CFLAG) -c src/ffasm.cpp -o obj/ffasm.o $(SFLAG) $(WINFLAG)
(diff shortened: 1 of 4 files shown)


Saffith:
There doesn't appear to be an issue in 2.50.x. There's no need for a compatibility rule if the problem doesn't exist in any released version.

ghost:
```
if((prvscr.ffflags[j]&ffSWAPNEXT)||(prvscr.ffflags[j]&ffSWAPPREV))
{
    int k=0;

    if(prvscr.ffflags[j]&ffSWAPNEXT)
    k=j<31?j+1:0;

    if(prvscr.ffflags[j]&ffSWAPPREV)
    k=j>0?j-1:31;

    zc_swap(prvscr.ffxdelta[j],prvscr.ffxdelta[k]);
    zc_swap(prvscr.ffydelta[j],prvscr.ffydelta[k]);
    zc_swap(prvscr.ffxdelta2[j],prvscr.ffxdelta2[k]);
    zc_swap(prvscr.ffydelta2[j],prvscr.ffydelta2[k]);
    zc_swap(prvscr.fflink[j],prvscr.fflink[k]);
    zc_swap(prvscr.ffwidth[j],prvscr.ffwidth[k]);
    zc_swap(prvscr.ffheight[j],prvscr.ffheight[k]);
    zc_swap(prvscr.ffflags[j],prvscr.ffflags[k]);
}
```

The above code from the 2.5.x branch says elsewise. It causes the same thing to happen actually it swaps everything when the tooltip says it should only swap speed.

Saffith:
It's the description that's wrong, not the behavior. In any case, it would make more sense to add a "swap speed only" checkbox rather than a quest rule.

ghost:
I would prefer that as well. I'll work on that today.

evouga:
Closing as there is nothing to do.

Feel free to send a new request with the "swap speed only" checkbox (if there's *really* a need for this?)

PR #28, opened 2016-08-08
Added OverlayTile(int, int) to ZScript.

ZoriaRPG:
This already existed in 2.50.x, but it was only ZASM. I added it to the
ZScript parser, and I _have_ tested the ZASM instruction (it works).


Saffith:
I don't want to add new functions and syntax to what's supposed to be a bugfix release. std.zh is sort of acceptable because it's separate from the program itself, but even that's pushing the limits of what I'm okay with.
Using 2.50.3 as a testing ground for 2.55 features is definitely not appropriate. That should be done in a 2.55 alpha.

ZoriaRPG:
My question, is: Is the oversight of OverlayTile not being in the script instructions table, but in every other table, not a bug? I mean, other instructions like GetGlobalRAM are in those tables, despite being useless, but an instruction that does work, and is useful, isn't listed. The quest output would still work in 2.50.0/1/2, too. 

Is your concern that people won't be able to compile a script that uses it in an older release? That IMO, is less of an issue, than modern scripts that rely on Link->Jump returning the correct value, which I ran into last week with Mitsukara. Such code would actually not work properly, but would compile. <shrug>

Did I miss anything, or add anything I should not have, for comment blocks, BTW?

Saffith:
A feature that doesn't exist, even if it was meant to, isn't the same as one that doesn't work right. And I'm getting really tired of the "other things are incompatible, too" argument. You can say that about practically anything.
It would just be nice if our version numbers weren't _totally_ meaningless.

ZoriaRPG:
_It would just be nice if our version numbers weren't totally meaningless._

I wholeheartedly agree with that. This is exactly why I was trying to plot out what versions marked specific amounts of change, but that seems to have gone to the wind, along with a lot of other things. I really didn't think we were going to do another 2.50.x update, and that we could bump the version ID, with changes that were appropriate to the bump.

The original idea, was that 2.53, would be a minor update, with the new drawing, with some fixes too. There could be subversions, but it would be a 'whole' number bump, followed by 2.54, and 2.55, which would be a new benchmark. More complex stuff would have gone into 2.60, and then even more, such as enemies, and weapons, changing, as 2.65. That was a two-year agenda, with all the major rewrites being incremental, leading to v3, which would be the full rewrite that you want. 

It was at least, something sane with a model/guideline for what the number actually means. Pretty much, anything 2.x, would still be Allegro 4, with ZScript. 3.x would be...well, whatever the plan evolves to mean. 

Do you have any proposed versioning scheme that is rational?

I would like to cut off 2.50.x after 2.50.3, and start working on 2.54/55. I'm intentionally skipping 2.51, and 2.52 (now 2.53) because users can;t keep straight that 2.50.3 would differ from 2.50.3. If you recall, 2.50.1 is commonly referenced as either 2.51, or 2.5.1, neither of which are appropriate. There's pretty much a guarantee that people would think that 2.51, the logical update to 2.50, is the same as 2.50.1. Avoiding user confusion, is pretty important, IMO. 

That is one reason to get off this 2.50.x path, and make clear that subversions, are exactly that. In fact, I'm in favour of using either a SR ID, or a Build ID to distinguish patch-y updates in the future. 

I'd also like to sort out what we all want out of each release, with some kind of reasonable roadmap. 

---

Anyhow, I'll eventually do some test builds with these new tidbits, and see how they play out. If you can roll the std.zh stuff into this, so that it's available, I'd appreciate it. I spent a lot of time ensuring that these things didn't have conflicts, and this does not include some of the things that I planned to do for 2.53, to keep it as close as what users expect, as possible. 

The majority of it, is completing sets of existing functions (and providing counterparts), providing very useful utilities, adding in the missing collision stuff, and more ways to handle collision and distance. 

A reasonable assortment of that was suggested, or requested by the primary scripters on Pure. 

Regarding the docs, whenever this seems to stall, and look as if it is going to drag on for another six months, I tend to taper off of working on them. 

Expect a comprehensive PDF for 2.45/55, and an updated set for this, later on. I really want to get some basic usage stuff, and tutorials in there, too. Realise, please, that people don;t know what _functions_ are, or what an identifier, token, or pointer _is_, as terminology. I intend to rectify that. 

I also need to revise the descriptions for all of the Link hitbox and draw properties, as none of them do anything, because (1) Link->Extend doesn't exist, and (2) even if it did, Link.cpp doesn't consider any of these variables when drawing Link. That's all on the to-do list, but zscript.txt needs to clarify that these do nothing. I _may_ already have done that, but IDR at the moment.  

---

How do you feel about about the error numbering inconsistency  (array errors), that I revised?

Saffith:
The error changes are fine.

Saffith:
I hadn't really looked at the .zh updates before. They... sure are a lot of stuff? I don't know about "very useful." Who needs a function to determine if a number is negative?
There are unimplemented functions, constants and comments that are misleading or incorrect, notes and commented-out code that just plain shouldn't be there... Once again, it's way too much stuff for two commits, and I'm inclined to toss it all rather than go through it to pick out the good bits.

Gleeok:
There is a lot of good stuff in there, however there is a lot of useless things as well.ex.;

int Lesser (int a, int b, int c, int d, int e, int f, int g, int h);
I've never had to use Min() for more than three values before. The exception is perhaps "int Min(int[] array, int count);"  -And "bool IsLess(...) is the same as bool value = (a < Min(...)); etc.."

Another one:
"bool DistY(int a, int dist)"  "Returns true if Link is with a distance of 'dist' pixels on the Y-axis, from the object specified."
-A Dist\* function that returns a boolean just has to be wrong. In general I think that "query functions" are always stupid. In practice something like "if (Distance(a, b) < 32) { ... }" can replace all of those.

Why don't we start a std2.zh (or a better named one) for future versions and let the scripting community help sort out the details?

ZoriaRPG:
Can we move this to the forum please? So much of this is a direct result of user requests, suggestions, and general feedback, that I feel that the userbase should be involved with the discussion.

I'm not really in favour of 'std2', and in fact, I prefer to include by default, and allow users to exclude what they want (see what I did with string.zh for an example). I would be in favour of splitting std_functions.zh into separate files, segregating the functions by type, or implied/intended purpose.

I'm going to defend DistX, DistY, and DistXY--although I could swap the identifiers with Proximity*, I went with Dist as it's shorter--as they are in fact tremendously useful. The same argument presented here could be made for Collision(), and these function similarly to Collision, save that they allow various sensitivities, and they are golden with things like LoS, and various quasi-collision and proximity effects where a differing distance on each axis is desirable.

Note that they are particularly useful because they accept all pointers, and combo locations, as the revised Collision() does, and I was going to add a param to measure based on other factors, including bool use_object_size, and bool from_centre.

I also caught a goof in my revised Distance(), that I fixed. I had forgotten an else statement, which is now fixed.

Lesser() and such were requested. They should be Min() overloads though.

Let's shift this to some soft of official forum topic though. None of the mainstream scripting userbase is reading this pull request topic...

ZoriaRPG:
If you're curious, the function to check if a value was negative or positive, was requested and submitted by ywkls. I don't get it either. He also needed the overloads for Min() and Max(), and I added the array versions because those, I thought would be extremely handy in the long term.

If you want me to go through and mark up what I personally added, versus what was resuested, suggested, or submitted,  so that you'll have a feel on the pulse of what the major scripters thought would be useful in the standard header, I'll do that.

evouga:
Closing this as stale.

I think many of these changes are fine (though agree with Gleeok that others are unnecessary) and am happy to pull them into the codebase. However,
 - they need to be applied to the (soon-to-be) master branch, not 2.50.x, as Saffith correctly states above;
 - they need to be broken down into individual additions, or small, closely-related bundles of changes, rather than one large set of ZScript changes.

Come to the forums and I can help sort things out.

PR #29, opened 2016-08-08
Updates Package (ZoriaRPG)

ZoriaRPG:
Added do_polygonr() and do_bitmapexr().
Crrected ffscript tables for them, and corrected DrawBitmapEx function to respect the valid number of args.
Added OverlayTile(int, int) to ZScript.
Added GetDMapScreenDoor(int, int, int) and SetDMapScreenDoor(int, int, int, int)
Added other getters for ZC interface conditions. 
Started adding new ZScript functions and ZASM instructions. 
Initial work on ItemData, and Link variables (Extend).
Added (at least partially) COMMENTBLOCK to the parser. 
Revised std.zh package, including std.zh (functions, constants), string.zh, added std_vars.zh, std.cfg, some shortcut routines, and other goodies. 
Additional, and revised docs.


Gleeok:
Is the script and drawing stuff based on the current trunk revision? -Because I'm pretty sure some of those would crash instantly.

ZoriaRPG:
I based it on the main trunk, from after you added the dummy functions for do_bitmapexr and do_polygonr.  We're still missing polygon3dr though, and rotation for polygonr. I pretty much abandoned an outline for polygonr, as the variable number of vertices makes using line() very impractical. 

I modified the pseudo-3d and 3d functions to allow using a bitmap ID as a texture: Is that the problem?

What precisely do you suspect would be crashy?

Saffith:
I see the DRAW_MODE_XOR stuff is in there again. I did explain several times that I don't want to add that because we cannot feasibly support it in the future, didn't I?

ZoriaRPG:
I don't need you to support it. I put them there, because I want to know if the bizarre issues with that mode, and the trans table carry over to bitmaps rendered with drawing trans sprites. 

I'm still retaining it in a personal fork, and just because it's in the file, doesn't mean that you need to retain it. I did alert you to this, back in January, and even there I agreed that it shouldn't become part of a mainstream release. (http://www.purezc.net/forums/index.php?showtopic=69279). I  can't feasibly maintain a fork, of a fork, either: At least, not without substantially more diddling with GitHub.

I'm still waiting for a response on _actual_ issues with the code, although I probably need to revise how I set up temp bitmaps for pseudo 3d things. I may have forgotten to get the subbitmap from the new resource pool...

ZoriaRPG:
I'll also address this in my next push, with a byte that's checked (bin flags) for allowing various modes that you maynot want to use, and I'll flag most of it off by default. I was going to tack on PATTERN_MODE and LIT_MODE, so I'll just do byte script_draw_flags, and set the params in the statements to check the flags, after evalluating the values.

Saffith:
Don't submit changes if you don't intend for them to be accepted. You altered 30,000 lines in a single commit. Do you really expect us to go over the whole thing and pick out which lines to keep?

ZoriaRPG:
Right... We've definately some issues to sort personally, and for now, what I'll do is clean up everything in an attempt to make it less work for you.

How do you want me to handle pull requests? I tried to do each file as an individual request, but GitHub decided to lump them together. Is there some way to force them to be separate?

Ultimately, if you're going to quibble about every minor thing to which you could possibly ever object, then reading commits--not just from me, but from anyone interested in contributing--is in the cards for you. At what point you'd decide to let go and relax about these things, is your choice.

Right now, you're still running this as if it was closed source, in spirit.

I need to know if the following draw modes will cause issues: Pattern, Lit.

Will we be able to support Lit mode, and is Pattern mode (extremely useful) going to pose a support issue? If you need a reminder, pattern mode fills all non-zero pixels with a specified colour, and IMO it should be safe.

---

P.S. @Gleeok: Could you please elaborate on what you see as an instant crash instruction? Is it the way that I added bitmaps to 3d drawing? I need to kn ow, as I'm planning to revisit this again, later this month.

evouga:
Closing as stale.

Zoria, once the master branch is fixed, please re-submit your changes one feature at a time. For example, send one pull request for OverlayTile, one for GetDMapScreenDoor, etc (and of course, make sure each works individually.) It may look like you're spamming us but actually it makes the job of reviewing and incorporating each change much easier.

PR #32, opened 2016-12-02
MSVC 2016 fixes and solution files

evouga:
Ported the MSVC 2008 solution files to work with MSVC 2016. Also updated a few source files so that they compile with MSVC 2016 (mostly issues with std::function and bind differing from their boost counterparts).


evouga:
Obsoleted by CMake.

PR #36, opened 2016-12-11
Define 'pack_fopen_password' in zcmusic.cpp & other fixes

AntumDeluge:
When I defined 'pack_fopen_password' in zcmusic.cpp, I was then able to compile without using the 'fixed' allegro distributed with the source.

The rest of the 'fixes' are small issues that I had dealing with objects not being created due to directories not existing. So I set a rule in the makefile to create the directories if they don't exist.

Another change was defining "ALLEGRO_UNIX=1" in alplatf.h if the "linux" rule was used, as it wasn't happening otherwise.

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>zelda-classic</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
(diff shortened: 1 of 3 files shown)


evouga:
The packfile fix is a good one but the Makefile changes are superceded by the CMake configuration. We will be looking at removing the need for a "fixed" Allegro (by upgrading to Allegro 4.4), but I'm closing this for now.

PR #43, opened 2017-01-16
Adds multiline comments. /* */

gwx:


evouga:
Can you explain a bit the regular expression `<COMMENT>([^*]|\n)+|.    { ; }`?

I'm not exactly sure what [^*] will match? Can this just be

```
<COMMENT>.     {;}
<COMMENT>\n    {;}

```

gwx:
[^*] matches "not *".
You could do that, but then you'd have to play around with ordering, maybe? I think generating a longer regex match is probably "faster" than treating each individual token as a null result.
I just pulled it straight from the example in the O'Reilly Flex/Bison book, so I assume it's the best way to do it.

evouga:
Ok. Works for me, then.

PR #45, opened 2017-01-17
Rearranged AST code.

gwx:
All the AST classes and case functions are now in the same order in the different files.

Also made a few functions const where it made sense to.


evouga:
Ok.

For future reference, "virtual" on overridden functions in child classes is OK (and documents that they are overrides) and don't need to be removed.

PR #50, opened 2017-01-21
Switch

gwx:
Adds a switch statement to ZScript.


evouga:
A couple of questions about this one:
1. How is the scoping handled here? I don't see any special code for it? For example what will the following code do? What **should** it do?
```
{
  int y = 50;
}
{
  switch(2)
  {
    case 1:
      int x = 10;
    case 2:
      Trace(x);
  }
}
```
It may be better to force a new scope for each case block.

2. What will happen if a user tries to "continue" from within a switch block?
```
while(true)
{
  // some array declarations
  switch(...)
  {
    case 1:
      // some more array declarations
      continue;
  }
}
```
My hunch is, nothing good?

gwx:
For the first, each case is its own ASTBlock, which I belive means each is its own separate scope.

For the second, it looks like it'll jump to the continue label, but only deallocate memory for the switch statement. (Which I realize now that I've done the string literals.) I'll go ahead and add in a separate continue ref thingy.

PR #51, opened 2017-01-23
Anonymous String Literals

gwx:


evouga:
I'm not really sure what the point is of addOpcode(), but the string stuff looks good

gwx:
My first attempt at the string code actually used addOpcode() for some fancier things, but I ended up not needing it in the end.

PR #52, opened 2017-01-23
Initial ZScript Reservations, and Added Itemdata

ZoriaRPG:
The majority of this, is the list of defines, and reservations in ffscript.h, ffasm.h, bytecode.h, and bytecode.cpp. All of that you should be able to pull with no consequences. 

The live code that is new, is Globalsymbols.cpp and ffscript.cpp, that adds in the itemdata stuff. 

I put the reservations changes in across several commits. I do not know if you can view them singly, or if github will lump them, but they are isolated, and necessary, as otherwise adding these things into this branch would be a disaster. 

String literal case values, and other bytecode stuff, I did not initially add all of this to the files in 2.future, at the end of those files, or in a sequential manner, but I _did_ do that, here. 

(The 2.future files have these definitions peppered around, but for this, I added them all at the end of the associated files.)

Again, these do not affect anything, unless you choose to use the ZScript that relies on them, but to maintain the greatest compatibility between canonical, and 2.future, it is prudent that any zscript labels have the same numeric values for assembly (registers, and such). This is the full list of ASM_DEFINE and SCRIPT_VARS that are in use, in the same order used for 2.future. 

The actual code that you might want to physically review, is in ffscript.cpp and GlobalSymbols.cpp. 

Maintaining the same function label IDs and variable IDs is important to me, so even if you never use them, please do not just delete them from the lists. Doing that would offset everything, horribly.

I did include all of the script drawing stuff that Gleeok is supposed to add, in the tables, and there are some nests of reserved values.

Anything that I plan to deprecate, I will change to a DEP* value in the tables, or a RESVD* value, so that it can later be turned into something useful with no offsets. Some things, such as the non-array values for idataflags, sprites, and attributes are still in the string literals, and in the tables, because I need to preserve ordering. but that are **not** in ffscript.cpp. 
	
My apologies if scanning through this seems like too much work, but adding in the IDs for these things a bit at a time is only going to create bugs, and otherwise generate oversights. I will be doing the other  zscript additions in individual stages, as I did for itemdata. You should be able to view these by commit though?


evouga:
This patch still seems like a mix of a lot of different things (including ZASM commands that have been specifically dis-approved here and on the forums).

It makes sense to preserve the numerical ID of ZASM registers, but this can be done without filling out the whole list at once?

Please focus on one feature at a time (for example, the screen transition effects WAVYIN etc) and commit only the changes relevant for this one feature.  We can start merging in the least-controversial features and go from there. Perhaps grayswandir can help you with structuring the patches?

ZoriaRPG:
One problem, is that I didn't add entries in any particular order, so to add in one instruction, I need to fill in the gaps between it, and the previous instruction. 

I didn't actually commit any of the unapproved functions, just the tables that list the instructions. THe instruction order needs to be applied at some point, and if it;s done in bits, then ordering the instructions so that they are committed in sets of like functions, would be extremely problematic. 

I suppose the opcodes and such don;t need to go in at one time, but the tables ASM_DEFINE, and script_command command_list; plus the ZASM Registers, and the Bytecode need to be committed in one go. There is simply no other way to preserve the order.

If I commit a partial list, and _someone else_ commits a change to this before I finish all the commits, that would offset the entire thing and completely wreck compatibility. 

When I submit the actual ffscript.cpp and global symbols functions, you could determine what you want to add, reject, or whatever, without ever caring what is in these other files. Unimplemented definitions do not harm anything. 

I wanted to submit they keyboard stuff next. To do that, everything between (#define READKEY 0x10F8) and (#define DMAPMAP 0x10B8)  in ffscript.h, and everything between (#define READKEY 408) and (#define DMAPMAP 333) in bytecode.h needs to be in place. 

If I want to submit the single function pair Set/GetDMapScreenDoor(), then the entire table needs to be in place, as it's essentially at the very end of the list. Submitting them in the order in which they appear on the tables is not at all practical, given your other requirements. 

It is far more practical to submit the live code, that you can accept or reject, in individual pieces. It makes it manageable, and it ensures that it is easier to comprehend what components are related. 

Further, that way, you can reject anything that you don't want to use, without that rejection impacting my fork, as well. The only viable alternative is that you 'cherry-pick' commits to my fork, but either way, if you don't want to ensure that a ZASM instruction in core isn't entirely different to a ZASM instruction in the fork, the tables need to be put in place, even if the instructions in them can never be called in main.

ZoriaRPG:
I can prune the opcodes if you want, but I'm not certain how much of a purpose that serves, other than to make organising them more chaotic. I submitted those in one pass to make it more clear which do what. 

You should be able to look through the individual commits though, and select those that you want to include.

ZoriaRPG:
Also, no-one is willing to work on structuring these patches. I've already tried to enlist.

evouga:
Hold on, let's discuss exactly what needs to be done to preserve the ordering:

1. The ffasm opcode and register list must include all entries in the correct order. Is it possible to leave the ones you are skipping as UNUSED** (the way that some are currently dummied out)? There is some work involved in dummying out registers, and the putting them back in later, but the main issue I want to avoid is a list of registers where it's unclear what is currently implemented, what is planned for the future, and what is in a broken half-implemented state. (Which I realize is a big problem with the *existing* ZASM instructions, but let's not make the problem worse.)

2. Everything ZScript-side can be #defined as needed while leaving gaps, right?

Gleeok:
The DrawBitmapEx function is over 1200(!) lines!? 

What do these do?
+ { "ARRAYSIZEB", 1, 0, 0, 0},
	+ { "ARRAYSIZEF", 1, 0, 0, 0},
	+ { "ARRAYSIZEN", 1, 0, 0, 0},
	+ { "ARRAYSIZEL", 1, 0, 0, 0},
	+ { "ARRAYSIZEE", 1, 0, 0, 0},
	+ { "ARRAYSIZEI", 1, 0, 0, 0},
	+ { "ARRAYSIZEID", 1, 0, 0, 0},

ZoriaRPG:
@Gleeok 

I set up DrawBitmapEx to accept a series of stacked flags to define its output styles. That took a whopping number of case values, but IMO it's for the best. All the various modes that it supports work best as a series of flags (trans, rotated, flipped), etc.

Those functions should be disabled. They return the _size of array_ for item, npc, itemdata, and so forth, as well as boolean arrays. All are proven to work, but I did not mean to include them. If they aren;t disabled, they should be. 

@DD: I can clip the defines, if that would make a difference overall. Just lots of extra copy-paste that seems redundant given that they will all need to be there, eventually.  The whole point was to prevent the need to go through and copy/paste stuff in and out, in each branch, as after a while of doing that, it just becomes confusing as hell, and hard to track, on my end. 

Is it hard to discern what isn't being used at this point?

The 'UNUSED*' values are registers that I deprecated. They can't just be removed, because that would also cause an offset. They can be reused though, in the future, as nothing is using them at present in any fork. I think they were some Link values that I converted?

Anyway, they were functions, and thus, can;t be trimmed, and the function order otherwise must be preserved.  When I add more functions, I will add them there, first. I believe that I trimmed those out in the last beta or so (?) freeing them for re-use. 

In fact, I already have an idea for how two, or all three of them will be recycled..

ZoriaRPG:
On the up-side, you should never need to add another bitmap function. There are no _possible_ modes for bitmap rendering that we can offer, that aren;t in there. (We clearly can't offer alphas, unless you want to do stacked rendering, 24b to 24b to 8b, to see if the trans results are any better; which is probably pointless, but I would of course be willing to test it...)

Once I finish the merge, and test all the disabled modes in 4.4, I will enable them. They needed inlines from libs that 4.2 just couldn't support... I had to cobble together a hacked 4.2/4.4 merge to get half of that to work, at all. 

There are of course mode combinations that were never implemented in allegro at all. I daresay that if i truly wanted, I could remedy that, in the future, but I probably shan't. unsupported mode couplings report an error of the specific nature of unsupported flagsets, when called, and I'll document them, of course.

Is there any way to optimise the zc_trans_table so that its colour selection is less perverse?

ZoriaRPG:
I'll also point out that I have a clear goal in preserving the ordering between zc main, and 2.future, so that registers used in 2.future _quests_ do not do something _radically different_ in main, in the event that I want to make a cross-encoder for them. People are already using 2.future, and I need to support them in the process. There's no reason not to preserve those offsets, even if you never include the functions/vars in canonical. 

I've done far too much work, and gathered too much interest to just kick my userbase in the trousers.

evouga:
What is butecode-new.cpp? Did you accidentally commit this? It doesn't look like it compiles?

There are many new registers still (most to do with with items). Can you briefly explain what these are intended to do?

What is the greyscale stuff? Does it actually work (when Link changes dmaps, etc)? Maybe move it into its own commit, since it seems unrelated to the item commands?

The reserved #defines do not bother me, with the understanding that their addition in no way signifies any commitment or approval to merging them into the source in the future. 

@Gleeok you added some drawing commands in historical/master that haven't been merged in yet. How do these relate to DrawBitmapEx?

ZoriaRPG:
What is butecode-new.cpp? Did you accidentally commit this? It doesn't look like it compiles?
_There are many new registers still (most to do with with items). Can you briefly explain what these are intended to do?_

They are the itemdata variables that were intended to go into 2.50, and were never added because people just left them up in the air. I'll push the itemdata section of zscript.txt for you. 

_What is the greyscale stuff? Does it actually work (when Link changes dmaps, etc)? Maybe move it into its own commit, since it seems unrelated to the item commands?_

Oh, the greyscale PALETTE iand function? They work, and should work across all dmaps, and dmap palettes. See: https://youtu.be/DUtVclNt5YU for a demonstration. It works across the entire system, and I debugged the hell out of it to ensure that it couldn't misbehave, but the actual ffscript functions that call it are not in the batch. I put the palette, and the palette function in there, but bot the rest of it, as I was going to submit those together with the screen transition effects.

_The reserved #defines do not bother me, with the understanding that their addition in no way signifies any commitment or approval to merging them into the source in the future._

I comprehend that, fully. There's no guarantee that anything that is committed, will be retained in the long-term, either. That's the nature of group projects, and why we have a separate fork for things that you don;t want to touch with a twenty metre cattle prod.

@Gleeok you added some drawing commands in historical/master that haven't been merged in yet. How do these relate to DrawBitmapEx?

evouga:
Ok, this is getting close. A few remaining minor issues:

ItemData.txt: 

there are still FlagX entries that need to be removed

bool Unused: let's get rid of this before some scripter gets "wise ideas" about its undocumented behavior

int Amount: this function is a mess (not your fault, but, still a mess): is there any to clean this up? Turn it into a function that separately sets the goal level and whether it is instant or drain?

int Max, MaxIncrement: these are also confusing. Is there a better way to control this behavior?


ffasm.cpp:

the following still needs to be dummied out: BITMAPEXR

ffscript.h:

there is some stuff in here that needs to be cleaned up (tempgreypal, some "ffrules" stuff, BITMAPEXR)

what is "LINKDIAG"? I think this was accidentally included.

ByteCode.cpp:

please remove the big dead code blocks


script_drawing.cpp:

there are still some changes in here. Is that intentional? What do they do?

ZoriaRPG:
I don;t think that any itemdata functions that predate my changes can be changed without the risk of breaking quest compatibility. Those functions work in the same general way as the item editor, in any case. I would be happy to work on them once I know how you set up reading the ffscript version, but changing them now is unwise.  

It isn't possible to change the variable names at all. They've been in use far too long. I do agree that they are poorly named. and I could add an overlay for them with a _second_ name. That's easy, but...not just yet. 25 hours on one hour of sleep, after another round of the same. 

I was planning to utilise the 'unused' flag, for 'Magic Cost is in _Seconds_', so sure, I can rip it out for now. 

LINKDIAG certainly should be disabled. 

I'll need to diff the scriptdrawing.cpp file. I think that the only other change was the patch for DrawInteger that I posted on the forums recently for you? I dropped it in there, when I added BitmapEx. , as it didn't look as if it was committed to the branch; and I forgot to disable it; unless you don't care about that one. 

It's probably three or four lines of changes for that function?

evouga:
If it's fixing the bug, it's fine. I just want to make sure the changes were intentional.

Gleeok:
Yeah, I'll make sure the drawing stuff gets in at some point so don't worry about it right now if that makes it easier. I'm actually happy that ZoriaRPG is working on them since I am a bit burned out on them atm (testing them, that is) is all, but we'll take them seriously.

ZoriaRPG:
_If it's fixing the bug, it's fine. I just want to make sure the changes were intentional._

That;s all that it was. 

_Yeah, I'll make sure the drawing stuff gets in at some point so don't worry about it right now if that makes it easier. I'm actually happy that ZoriaRPG is working on them since I am a bit burned out on them atm (testing them, that is) is all, but we'll take them seriously._

DD had me pull DrawBitmapEx because he thought you had a problem with it? Was it just that the size of the function shocked you, or was there some issue with it? It's long, because it has a huge list of cases for output mode flags, with four primary modes of each (scaled, not scaled, masked, not masked). 

It works very well from a user-perspective though. DrawBitmapEx(... MODE_TRANS+MODE_PIVOT...) and such.

evouga:
Almost there! Now just down to a few typos:

All files:
"RESERED" should be "RESERVED"

ItemData.txt:

This is not needed:

//! New -Z

since it will soon be out of date. But feel free to note in the commit log which functions are new (so that it is easy to collate the change list for the next release announcement)

ffasm.cpp:

FREEx00F3 -> FREE0x00F3

ffscript.h:

BITMAPEXPR still needs removing

ZoriaRPG:
Which file is the commit log?

evouga:
Oh, I just meant the message that you enter as part of a `git commit` (not a specific file).

ZoriaRPG:
Oh.

I also keep a change log separately, for this sort of thing.  Pushed the corrections.

ZoriaRPG:
I also committed a tiny bugfix to do_drawbitmapr(). The previous versions have a constraint to a height and width of 513, not 512.

evouga:
Looking good. One last thing: there is still a big block of dead code in ByteCode.h, can your remove it?

ZoriaRPG:
Done.

evouga:
Looks good. Thanks!

PR #53, opened 2017-01-25
Added a basic reference sheet for working with weapons.cpp and weapon…

ZoriaRPG:
This is a reference sheet for class properties and functions, related to Link weapons. I don't know if you want to include things of this sort. If not, just close the pull.

### DIFF
--- a/Docs/Ref_Items_LinkWeapons.cpp
+++ b/Docs/Ref_Items_LinkWeapons.cpp
@@ -1,0 +1,256 @@
+// Link Items and Weapons Reference
+
+
+/* Item Class Variables
+	The following are the present values that can be passed to items in weapons.cpp or link.cpp, 
+	from the item editor in ZQUest.
+	All examples use 'curitem' as their pointer, for example. curitem.family. 
+	They vurther use a legal assign, as an example.
+*/
+
+int var;
+
+//Item Class, ->Family in ZScript
+var = itemsbuf[curitem].family;
+
+//Level, ->Level in ZScript
+var = itemsbuf[curitem].fam_type;
+
+//Keep old, ->Keep in ZScript
+var = itemsbuf[curitem].flags & ITEM_GAMEDATA;
+
+//The 'Increase Amount' value in the item editor, ->Increase in ZScript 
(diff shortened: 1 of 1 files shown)


evouga:
This looks like very useful documentation, but I don't understand why it is in a separate file (and a .cpp file, especially). Shouldn't this documentation go into zdefs.h, integrated into struct itemdata, etc?

ZoriaRPG:
I, and other coders tend to find it useful to have refs out and open as separate files, so that you don't need to skim through the source to find them. It also covers variables used in at least four different parts of the codebase, so this is a central way to view it. if you don;t want to include it, that;s fine. I'll just put them all on pastebin, which I was also doing for the ZScript code docs.

I made it a .cpp file so that it syntax highlights functions, variables, and the like, for easier reading. I find that easier anyway, with bad eyesight. 

I was going to make another one of these each, for npcs, for Link, and for eweapons, so that people new to the project can read them and tell what all these vars and functions do, at a glance, plus of course the one for ffscript. Making one for the core zelda and zquest things is something I'd consider doing in the future. 

At least in regard to documentation, I do have official standing, and I was _supposed_ to make all of these, eventually, but there is no mandate that they need to be on GitHub. It;s really trivial to me, and I was going to include it with the source to make it easier on people new to it, to delve into the codebase and avoid getting perpetually lost. 

Cluttering the source with in-line instruction manual content, is also extremely messy, and the worst possible approach in my opinion, as a technical writer. 

I'm not going to go back and forth on this stuff and endlessly discuss its contents, or format, _at present_. if you don;t want to include it, just close the pull, and I won't push more docs, other than the ZScript/ZQuest and similar text files, to GH. Documentation in _any form_ is useful, at this point, IMO, and until I have a sufficient quantity of it, having a group debate on its packaging is just banter. 

I can barely see _at all_ right now, so aye, syntax highlighting is why I did it the way that I did it.

evouga:
Ok. I don't think it makes to add the documentation in this form to the main repository, but feel free to submit pull requests for better documentation of the existing source files (e.g. zdefs.h).

PR #54, opened 2017-01-26
Closed: Added Visual Effects, also Fixes; now in a newer pull that has been merged. 

ZoriaRPG:
This pull adds in the following visual effects:
Monochome Mode (On, Off)
Wavy Transition (In, Out)
Zap Transition (In, Out)
Opening Wipe Transition

These modes were fully-tested in 2.54, with extensive checking that the monochrome palette wasn't improperly set, or improperly unset. The logic for it is primarily in pal.cpp, and all palette effects must be handled therein. 

It also fixes the ASM_DEFINE, ZASM Registers, and related arrays, correcting identifiers, and identifier pairs that I missed previously. This required changing the length of an instruction label from 16 to a larger value, to compile. I set the instruction name size to [64] for now. 

Is there any issue that changing this size will create? 

I removed the /Docs path. I don;t know if this is why github can't automerge, or if something else is causing this issue. Perhaps one of you can determine why automerge isn't possible.


evouga:
You need to pull master and merge it into your branch.

This occurred because I already pushed some fixes to your previous commit.

evouga:
You need to correct the conflicts before I can merge these changes. The screen effects seem trivial, but I still have a question about the monochrome mode: where does tempgreypal first get set?

PR #55, opened 2017-02-04
Fix compiling on linux.

gwx:


Gleeok:
No! c++ headers like algorithm are like 40,000 lines of crap! Also don't put it in the header file when you only need it in the source file!!! Use zc_min, or whatever trivial thing will do in the future! 

:)

gwx:
Yeah, I really have no idea what to do here, aside from just rewriting it to avoid min/max completely. If I try defining min/max I'm liable to break it on the windows side. I'll take a look at zc_min, though.

PR #58, opened 2017-02-11
Type objects

gwx:
Adds ZVarType and its subclass, ZVarTypeSimple. SymbolTable will store a copy of each unique type, and allow you to convert to and from types and type ids.

The last commit is the bulk of the work. The others were just cleaning things up so I could do the last one without freaking out too much.


evouga:
Wow. I assume this is all leading up to adding structs as "complex types" to the type system.

PR #109, opened 2017-07-02
2.53 / 2.50.2.1 Changes

ZoriaRPG:
These are the changes to 2.50.2.1, or 2.53, re-implemented. As far as I can tell, this should be effectively identical to the previous revision that I posted as a separate trunk. Please try a pull+compile to see if I missed anything minor that has otherwise changed.


evouga:
It looks like you forgot: https://github.com/ArmageddonGames/ZeldaClassic/commit/c5c9c9bca992c99cff688b7ab56298f1ad84d927

PR #116, opened 2017-07-14
2.53: Fixed ag.cfg Issues

ZoriaRPG:
This fixes the corruption of ag.cfg that occurs when ZQuest closes.


evouga:
This patch has some redundant fixes try that were already incorporated (ag.cfg), unexplained changes (to quest rules, joystick, etc), undoes previous fixes without explanation (midi loading).

PR #118, opened 2017-08-20
Fix: Lens Script Stealing

ZoriaRPG:
Changelog: Fixed a bug where the Lens of Truth steals the script ID from
the last (other) item used, and runs that script every frame until the
lens is suspended.


evouga:
ok.

Issue #117, opened 2017-08-15
Linux build process not working as expected


ntfwc:
Following the steps as written in the readme:
```
cd build
cmake ..
make
```
`make` fails with `fatal error: allegro/platform/alunixac.h: No such file or directory`

Copying alunixac.h from allegro/extra to allegro/include/allegro/platform allows it to build. But the resulting zelda binary does not have sound, and looking at it with ldd reveals that, for some reason, libasound is not linked.

On the 2.50.2 build from the website, sound does work and it has libasound linked.

I did try getting a fresh copy of allegro 4.4.2 to generate an alunixac.h. The result did define a few alsa variables the other did not. But copying that to allegro/include/allegro/platform still gave me the same result.

Linux Version: Xubuntu 16.04 64-bit
Commit: c8701fc9320a873e9542483c11398272ac0ccdad

evouga:
How did you compile Allegro? (You may have to tinker with the configuration to enable sound support? You can try asking at the Allegro forums.) Does sound work in any application using that build of Allegro?

evouga:
You can also try compiling the various sound libraries from source (in the other/ directory) rather than using the provided binaries.

Sorry I can't help more; the only Linux machine I use has never been able to get sound to work with ZC. The developer who compiled the 2.50.2 binaries is no longer actively involved with development, and the rest of us use Windows.

ntfwc:
That's unfortunate. To clarify, I didn't compile allegro, I just let ZC use the pre-compiled library from the git repo. I can try building it and make another attempt.

ntfwc:
So I looked at it over the weekend. I did succeed in building allegro with alsa, but I ended up figuring out that I don't actually need alsa for the audio to work.

It turned my sound problem stemmed from the fact that I was depending on it to pipe audio through pulseaudio, which, without alsa, is not what is done. Sound does work if I run it from my main desktop user account, which, from what I gather, means it has to be run from an account with direct access to the sound device. So it shouldn't be a problem in most cases.

Everything but midi music is working, which I investigated and I'll add a different issue for.

I have added a pull request with a fix for alunixac.h

andlabs:
Should this fix have been merged into the 2.55 branch (default)? It's failing for me too without it.
EDIT Manually patching it in doesn't fix it; in fact alunixac.h isn't even generated...

PR #120, opened 2017-08-21
Fix for #117

ntfwc:
This allows ZC to build without issue on Linux.

### DIFF
--- a/allegro/CMakeLists.txt
+++ b/allegro/CMakeLists.txt
@@ -429,6 +429,6 @@ endif(WANT_LINUX_CONSOLE AND CMAKE_SYSTEM_NAME STREQUAL Linux)
 if(UNIX)
     configure_file(
         ${CMAKE_SOURCE_DIR}/include/allegro/platform/alunixac.h.cmake
-        ${CMAKE_SOURCE_DIR}/extra/alunixac.h
+        ${CMAKE_SOURCE_DIR}/extra/allegro/platform/alunixac.h
         )
 endif(UNIX)


evouga:
Isn't this file generated by Allegro's configure scripts? If so maybe it's better to invoke Allegro rather than add a specific instance of the file? (I thought I had set something like that up for Linux compilation but it could be broken...)

gwx:
I cloned a clean repo to try building, and I didn't encounter any error. Trying with the fix seems to work fine as well, so I'm not quite sure what's going on here.

ntfwc:
For me, if I build clean without this change, I get
`
fatal error: allegro/platform/alunixac.h: No such file or directory
`
So I don't know, maybe there is another instance of it somewhere in your include paths?

evouga:
Here is the part that should invoke that Allegro config script: (CMakeLists.txt line 150)

`execute_process(COMMAND "${CMAKE_COMMAND}" "-H${CMAKE_SOURCE_DIR}/allegro" "-B${CMAKE_SOURCE_DIR}/allegro" WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/allegro")`

This should invoke allegro/CMakeLists.txt, and in particular run the following bit at the end

````
if(UNIX)
    configure_file(
        ${CMAKE_SOURCE_DIR}/include/allegro/platform/alunixac.h.cmake
        ${CMAKE_SOURCE_DIR}/extra/alunixac.h
        )
endif(UNIX)
````

which should generate the include file.

Is there anything in your CMake output that could tell us which part of the above isn't working correctly on your machine?

evouga:
Actually I think I see the issue. Can you try the latest commit?

ntfwc:
The latest commit builds fine

PR #125, opened 2017-08-27
Fixed alleg_compat methods to cleanup the packfile password. Closes #121

### LINKED ISSUE #121: Midi files are not playing
This affects both zelda and zquest. In zquest, if you try to play music from the Etc>The Travels of Link menu, nothing happens. Same result if you try to play a midi from a quest through Quest>Audio>MIDIs. In zelda, if you play a quest with midi music, the music does not play.

This problem does not occur in 2.50.2.

Investigation:
I was able to bisect the problem and found that the last good commit was 1ca34af68d3c009201e70373e70416dbb0c7260c, and the first known bad commit is f95b80061749ac4ead6fb1cbf76660b2e5540c6c (as the previous commit could not be built). Combing through the changes I discovered that this was introduced with the removal of two methods, file_size_ex_password and pack_fopen_password, from file.c and the introduction of alleg_compat.cpp. So there seems to be some problem with alleg_compat.cpp's re-implementation of these methods.

OS: Xubuntu 16.04 64-bit
commit: c8701fc9320a873e9542483c11398272ac0ccdad

ntfwc:
This should restore the old behavior, allowing samplesoundset/patches.dat to load

### DIFF
--- a/src/alleg_compat.cpp
+++ b/src/alleg_compat.cpp
@@ -11,10 +11,14 @@
 
 PACKFILE *pack_fopen_password(const char *filename, const char *mode, const char *password) {
 	packfile_password(password);
-	return pack_fopen(filename, mode);
+	PACKFILE *result = pack_fopen(filename, mode);
+	packfile_password(NULL);
+	return result;
 }
 
 uint64_t file_size_ex_password(const char *filename, const char *password) {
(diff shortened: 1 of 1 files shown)


Gleeok:
Thank you! We've just added it. :)

PR #126, opened 2017-08-27
2.50.x tile editor fix

ZoriaRPG:
Fixed an issue where the tile editor refused to load images (PNG/JPG) on the first attempt.


ZoriaRPG:
The same patch needs to be applied to master. My master branch is out of sync, in ways that require me to force replace it, and I need to archive all of my changes before I can do that safely. 

I would appreciate it if someone would apply this change to master as well, in the meanwhile.

PR #134, opened 2017-09-04
Refactor: Link and Doors

ZoriaRPG:
Changelog: Refactored Link;s door interactions, to separate diagonal
and orthogonal checks.

Note, my previous fix did not do the trick, nor does this, but it is required, and reasonable, to split these up, 
so that a proper fix is possible. I now see and understand the issue that was reported quite clearly, and the door checking needs to poll input, when in diagonal movement mode, rather than check direction. 

Link's direction and his tile are not always a match, and the doors do not open when you would expect them to open. Link can glide back and forth along walls, and the door will not open until he stops, and presses a direction again. Thus, I have split diagonal and orthogonal checks for doors into their own sections, and I will follow this up with a fix for the actual issue.

### DIFF
--- a/src/link.cpp
+++ b/src/link.cpp
@@ -8710,17 +8710,401 @@ void LinkClass::checklocked()
     int si = (currmap<<7) + currscr;
     int di = nextscr(dir);
     
-    /*
+    
 	
 	if ( diagonalMovement ) 
 	{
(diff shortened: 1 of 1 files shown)


evouga:
Fine, though it's not entirely clear why the logic cannot be unified.

evouga:
Also, please apply all fixes to both 2.50.x and master.

PR #135, opened 2017-09-07
Fix: Diagonal Link NES Doors

ZoriaRPG:
Changelog: Fully implemented a fix for opening NES sungeon doows while wall
walking with diagonal movement enabled, to match the behaviour of lock blocks.

**Do we need a quest rule for this behavioural change, under compatibility rules?**


evouga:
1. Instead of

````
if( !diagonalMovement && pushing!=8) return;
	/*This is required to allow the player to open a door, while sliding along a wall (pressing in the direction of the door, and sliding left or right)
	*/
if ( diagonalMovement && pushing < 8 ) return; //Allow wall walking Should I add a quest rule for this? -Z
````

Can it just be `if(pushing < 8) return;`? Is there a need for different behavior in the diagonal vs non-diagonal movement behavior?

2. What is the purpose of this code?

````
bool found;
	for ( int q = 0; q < 4; q++ ) {
		if ( tmpscr->door[q] == dLOCKED || tmpscr->door[q] == dBOSS ) { found = true; }
	}
	
if ( !found ) return;
````

3. Please do not leave in blocks of dead commented-out code (beginning line 9092).

ZoriaRPG:
1. This would potentially change the behaviour of how doors work in 4-way movement. I do not know if it could cause issues, so I only added a new check on pushed for diagonal. I initially did simply make it a check against pushed < 8, but given that the check was == 8 for NES doors in the past, and then I decided not to change true classic behaviour. It's an easy thing to convert, but I would want user feedback on that change.

2. It exits if there are no locked doors on the screen, so that the function ends without all the directional and positional checking in that instance. (Performance). The checklockblock() stuff has a similar performance check, IIRC. 

3. Fixed. 

I'll leave it without a QR for now. if there are user complaints, then we can address them later.

evouga:
1. What I mean is, under what circumstances is there a difference between the current pair of checks, and looking for pushing < 8? The only thing I can think of would be, for example: Link has no keys and is pushing up against a locked door. A secret triggers and drops a key directly onto Link. In the current code, Link must stop pushing and start pushing again to unlock the door. In the simplified check, Link would automatically unlock the door.

In other words, I want to make sure the current behavior is correct/desirable; there's no use complicating the code in order to perpetuate a buggy behavior.

2) Fine, though it seems unnecessary.

ZoriaRPG:
I'm not sure. We would need to check if the actual NES behaves in this manner. If it does, then it is a NES consistency concern. If not, then what we were doing is simply erroneous, and we can convert it. 

I don't think that the bug can occur with 4-way movement though, as as soon as Link touches a door, pushing starts. Link cannot wall walk in 4-way movement. I suppose I can simply remove the conditions and set pushing >= 8 in general. I'm mostly wary of changing very old behaviour without a bloody good reason, while the fix only affects more modern behaviour that has been buggy for ages. 

_i.e.,_ I do not want to add complex behavioural checks that we will need to patch with a QR later, if they aren't useful. 

I do sort of want to split all Link checks into 4-way and 8-way, at some point, so that there is a clearer flowchart for them. There is so much 8-way code that is flawed, that it may as well be a separate case.

evouga:
Please send a fix for master branch as well.

ZoriaRPG:
Patched along with all of the others.

PR #154, opened 2017-10-02
Chore: More patches. Solidity, CSet, and Data seem Good. 

ZoriaRPG:


ZoriaRPG:
What the hell? If I force-quit ZC, everything works perfectly, but if I quit ZC normally, and reload it, then it starts to crumble? Why are all of these functions so broken???

PR #166, opened 2017-10-06
Segregate ag,cfg and fix ZQ-> Clear File Path

ZoriaRPG:
The configuration file (ag.cfg) is now two files, zc.cfg and zquest.cfg. This should aid in preventing both programmes from sharing access tot he same file, and thus potentially overwriting settings. 

Further. this fixes the implementation of the ZQuest menu option Çlear Quest Filepath. to remove the default filepath in ZQ, after moving the location of the binaries.

### DIFF
--- a/src/zelda.cpp
+++ b/src/zelda.cpp
@@ -2869,9 +2869,9 @@ int main(int argc, char* argv[])
     loadpng_init();
     
     // set and load game configurations
-    set_config_file("ag.cfg");
+    set_config_file("zc.cfg");
     
-    if(exists("ag.cfg") != 0)
+    if(exists("zc.cfg") != 0)
     {
         load_game_configs();
     }
--- a/src/zquest.cpp
+++ b/src/zquest.cpp
@@ -21416,8 +21416,8 @@ int main(int argc,char **argv)
     jpgalleg_init();
     loadpng_init();
     
-    set_config_file("ag.cfg");
-    
+    //set_config_file("ag.cfg");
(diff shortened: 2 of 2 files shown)


evouga:
Which one does Romview read from / write to?

ZoriaRPG:
@DarkDragon

ROMView writes to ag.cfg, which also contains more generic values, and values for ZLaunch;l so you need not worry about updating the RomView source at present. The allegro (ag) config file can be preserved for this type of data application, as some values should not present a conflict:

The special sections used by ZCL (and only by ZCL) also write to ag.cfg, and graphics drivers writes to all three, for the present, as the drivers used by allegro should be uniform on most systems. (This is a ZCL override.)

We'll see if there are any issues in some tests. I can always modify the ZCL behaviour further, if desired, or needed. 

ZCL Source and binary output is here for the present:
https://github.com/ZoriaRPG/ZCLaunch

Nice to see that you are back, BTW. We've missed you. :)

Gleeok:
If these are truly to be split up I'd rather see only a zc.cfg, zq.cfg, and rv.cfg. If there is still state sharing between them then why not keep a single ag.cfg?

ZoriaRPG:
The only crossover is from ZCL. I didn't want to much with ROmView to change its config file at present, as it has other issues that are a higher priority on my list.

I could add zcl.cfg too, which would take care of the remnant crossover. 

The actual ZC programmes cannot read or write across files: Only ZCL does this, and only to set some gfx modes.

PR #191, opened 2018-04-03
Enemy::Lined_Up fix

ghost:
fix 8dir homing being way too sensitive.


ghost:
Moving this into a separate branch... called 2.54 enemies

PR #203, opened 2018-09-23
Remove inclusive ORs left over from QR porting

EmilyV99:


ZoriaRPG:
All of the inclusive OR removal needs to be:

|| ( FFCore.getQuestHeaderInfo(vZelda) < 0x192 && (get_bit(quest_rules,qr_QUICKSWORD) )

...rather than plain removal, as in builds prior to 1.92, there is no section version to use, and the item bits will never be set.

PR #231, opened 2018-10-23
2.55 tintdmapfix

EmilyV99:
Zoria, please double check this if you could


ZoriaRPG:
I made a backup of 2.55, and I'll pull it. One thing, is that the monochrome function might be used in legacy code, for when Link dies. 

We may need a separate routine for the scripted 'greyscale' effect, because the old one had very specific behaviour with interpolated fades.

PR #302, opened 2019-03-22
Namespace

EmilyV99:
`namespace NAME {CODE}` where the CODE can include functions, scripts, variables, constants, and options. Namespaces are not tied to FileScope, and thus have their own options!
Added SCOPERES operator ("::"), used exclusively in ASTExprIdentifier for namespace resolution.
	`namespace::var` is valid. `script.var` is valid.
	`namespace.var` is invalid. `script::var` is invalid.
Added several bison nonterminals, allowing now `Identifier_List` (containing any list of identifiers) as before, as well as `Scoperes_Identifier_List` which cannot contain `.`, and (commented out as it is currently unused) `Dot_Identifier_List`, which cannot contain `::`.
Added a way to declare multiple namespaces at once: `namespace name::name2{}`. This declares a namespace `name`, then declares `name2` within it.


EmilyV99:
Bug found, crash-causing level. Don't merge this yet.

EmilyV99:
Fixed the crash in the using commit. Should be all good to merge!

PR #314, opened 2019-03-31
Options dialogue + Short-Circuit

EmilyV99:


ZoriaRPG (commented):
Aye, you pretty much always want GOTOTRUE for this.

PR #324, opened 2019-04-08
Add itemsprite Scripts to the Engine

ZoriaRPG:


Alucard648:
I am guessing, it`s for dropped items to control their movement and behaviour, such as dropped fairies.

PR #385, opened 2019-06-25
Feat: New flippers options

EmilyV99:
Attribute 1: Flipper dive time
Attribute 2: Flipper cooldown time
Flag 2: Cancellable dive
Updated V_ITEMS to 45
Replaced various `diveclk>30` calls with `isDiving()` calls, which previously just returned `diveclk>30`, but now handles the new options.
Included sanity checks for this at several points.
If no flippers are had, the default values of 50 (dive) and 30 (cooldown) will be used.


ZoriaRPG:
Missing initialisation of these Misc values to values reflecting their previous hardcoded defaults in qst.cpp for V_ITEMS < 45 in readitems()

ZoriaRPG (changes requested):
Missing initialisation of these Misc values to values reflecting their previous hardcoded defaults in qst.cpp for V_ITEMS < 45 in readitems()

PR #392, opened 2019-06-30
Parser: `MAX_INT_ONE_LARGER` now `TRUE_INT_SIZE`

EmilyV99:
This allows `214748.3647`/`-214748.3648`
All appropriate constants in std_constants have been updated.
Also, `ftoa` has been patched to not bug out on constants of `MIN_CONSTANT`. A variable of MIN_CONSTANT will potentially still have issues, but those were also present before this commit.

### DIFF
--- a/output/common/std_zh/std_constants.zh
+++ b/output/common/std_zh/std_constants.zh
@@ -25,17 +25,17 @@ const float SQRT_MAX			= 463.4095; //The largest square root thet ZC can return.
 ////////////////////////////
 /// ZC Minimums and Maximums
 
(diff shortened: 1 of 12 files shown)


EmilyV99:
This will require a block to be copied to RegistrationVisitor in that branch, once this is merged.

ZoriaRPG (commented):
The value range is 214748.3647/-214748.3648???

It should be consistent. Does positive 214748.3648 break?

EmilyV99:
That range is the true 32-bit range. The only issue is attempting `Abs(-214748.3648)`, as that gives back `-214748.3648`; but this issue already exists with variables.

PR #395, opened 2019-06-30
CompileOption: Binary 32-bit operations/literals

EmilyV99:
With this compile option on, binary literals and operators will be direct (skipping `/10000` and `*10000`)
This allows the use of 32 bits per variable.
This option is forced to OFF in `std.zh`.
This option defaults to OFF.


ZoriaRPG (commented):
I merged this, but I want a quest bit to allow toggling this ad-hoc, and to set it as a default compiler setting.

PR #405, opened 2019-07-11
Fix: A number of MSVC compiler warnings. A large number.

EmilyV99:
Various `unused local variable`, `cast to bool`, and `signed/unsigned mismatch` warnings.


EmilyV99:
Closing for: Too many conflicts.
This can be re-done from scratch if needed.

PR #456, opened 2019-10-01
Fix: Allow overwriting LA_HOPPING

EmilyV99:
THIS NEEDS A COMPAT QR STILL.


EmilyV99:
still needs compat QR... should do this very soon

PR #533, opened 2019-12-13
Feat: Wrapper for running scripts

EmilyV99:
sprite::run_script(int mode) allows running sprite scripts directly
	from a sprite pointer, or from a sprite_list for all members.


ZoriaRPG (changes requested):
waitdraw = false;

Waitdraw on sprites is a byte. This should flag a performance warning in MSVC. The correct assigns are 1/0.

PR #562, opened 2020-02-29
Giving cursor in Mooshmood theme a natural color instead of red

Rekiron:


EmilyV99:
Submit as separate theme if desired.

PR #594, opened 2020-04-04
Alpha now 58, includes the prior two commits for GetPixel.

ZoriaRPG:


EmilyV99:
This was forgotten, we already bumped version.

PR #638, opened 2020-08-18
Standard Conformance, level 3 warnings, and const rules

SyrianBallaS:
This PR was tested using VS 2008, ZeldaClassic has a significant performance boost. Never dropped frames even in debug mode.

For ZQ, all that was tested was script compile, moving tiles around, saving, exiting. If there is other things that need testing let me know. Everything should work.

A `config_compat.h` file was made to not break compatibility with the current tootset. 

ffscript.ypp was mostly hard to follow, some the code could be analyzed for `const` variables but most could not, there about 48 warnings about non-class `const` rules but are ignored for now. CMakeList.txt was not touched.

There should a be a way to add the `-permissive` switch in `CMakeList.txt` for all supported compilers but will do that later.

Review comments will be added after this PR is created.

### DIFF
--- a/allegro/include/allegro/inline/fmaths.inl
+++ b/allegro/include/allegro/inline/fmaths.inl
@@ -57,7 +57,7 @@ AL_INLINE(double, fixtof, (fixed x),
 
 AL_INLINE(fixed, fixadd, (fixed x, fixed y),
 {
-   fixed result = x + y;
+   const fixed result = x + y;
 
    if (result >= 0) {
       if ((x < 0) && (y < 0)) {
@@ -80,7 +80,7 @@ AL_INLINE(fixed, fixadd, (fixed x, fixed y),
 
 AL_INLINE(fixed, fixsub, (fixed x, fixed y),
 {
-   fixed result = x - y;
+   const fixed result = x - y;
 
    if (result >= 0) {
       if ((x < 0) && (y > 0)) {
--- a/allegro/include/allegro/inline/gfx.inl
+++ b/allegro/include/allegro/inline/gfx.inl
@@ -101,7 +101,7 @@ AL_INLINE(void, clear_to_color, (BITMAP *bitmap, int color),
 })
 
 
-AL_INLINE(int, bitmap_color_depth, (BITMAP *bmp),
+AL_INLINE(int, bitmap_color_depth, (const BITMAP *bmp),
 {
    ASSERT(bmp);
 
@@ -109,7 +109,7 @@ AL_INLINE(int, bitmap_color_depth, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, bitmap_mask_color, (BITMAP *bmp),
+AL_INLINE(int, bitmap_mask_color, (const BITMAP *bmp),
 {
    ASSERT(bmp);
 
@@ -117,7 +117,7 @@ AL_INLINE(int, bitmap_mask_color, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_same_bitmap, (BITMAP *bmp1, BITMAP *bmp2),
+AL_INLINE(int, is_same_bitmap, (const BITMAP *bmp1, const BITMAP *bmp2),
 {
    unsigned long m1;
    unsigned long m2;
@@ -135,7 +135,7 @@ AL_INLINE(int, is_same_bitmap, (BITMAP *bmp1, BITMAP *bmp2),
 })
 
 
-AL_INLINE(int, is_linear_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_linear_bitmap, (const BITMAP* bmp),
 {
    ASSERT(bmp);
 
@@ -143,7 +143,7 @@ AL_INLINE(int, is_linear_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_planar_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_planar_bitmap, (const BITMAP* bmp),
 {
    ASSERT(bmp);
 
@@ -151,7 +151,7 @@ AL_INLINE(int, is_planar_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_memory_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_memory_bitmap, (const BITMAP* bmp),
 {
    ASSERT(bmp);
 
@@ -159,7 +159,7 @@ AL_INLINE(int, is_memory_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_screen_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_screen_bitmap, (const BITMAP* bmp),
 {
    ASSERT(bmp);
 
@@ -167,7 +167,7 @@ AL_INLINE(int, is_screen_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_video_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_video_bitmap, (const BITMAP* bmp),
 {
    ASSERT(bmp);
 
@@ -175,7 +175,7 @@ AL_INLINE(int, is_video_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_system_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_system_bitmap, (const BITMAP *bmp),
 {
    ASSERT(bmp);
 
@@ -183,7 +183,7 @@ AL_INLINE(int, is_system_bitmap, (BITMAP *bmp),
 })
 
 
-AL_INLINE(int, is_sub_bitmap, (BITMAP *bmp),
+AL_INLINE(int, is_sub_bitmap, (const BITMAP *bmp),
 {
    ASSERT(bmp);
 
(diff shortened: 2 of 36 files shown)


allegro/include/allegro/inline/fmaths.inl
```diff
@@ -57,7 +57,7 @@ AL_INLINE(double, fixtof, (fixed x),
 
 AL_INLINE(fixed, fixadd, (fixed x, fixed y),
 {
-   fixed result = x + y;
+   const fixed result = x + y;
```
SyrianBallaS, line 60: Most of the changes is just adding `const` where it should be for variables and parameters. `const` correctness for classes/structs was not checked for this PR.

allegro/include/allegro/inline/gfx.inl
```diff
@@ -101,23 +101,23 @@ AL_INLINE(void, clear_to_color, (BITMAP *bitmap, int color),
 })
 
 
-AL_INLINE(int, bitmap_color_depth, (BITMAP *bmp),
+AL_INLINE(int, bitmap_color_depth, (const BITMAP *bmp),
```
SyrianBallaS, line 104: Passing pointer to `const` is important, not only is it a performance boost, it makes it easier for maintainers to know for sure it should not be `free`'d in this scope. 

Later (way later) we could benefit by adding a `restrict` qualify to pointers if there is more than one pointer parameter.
Requires knowledge of the strict-aliasing rule to use correctly.

Example
```
memcpy(void* restrict dest, const void* restrict str, size_t count);
```
You can read more about the `restrict` qualifier (C99) here: https://en.cppreference.com/w/c/language/restrict
Most compilers offer at as an extension (gcc, vc++, clang)
Just something to keep in mind for later.

src/compat_config.h
```diff
@@ -0,0 +1,31 @@
...
+#elif defined(__GNUC__) || defined(__clang__)
+#define _MAYBE_UNSUED(x) x __attribute__((unused))
+#define _FALLTHROUGH __attribute__((fallthrough))
+#elif defined(_MSC_VER)
+#define _MAYBE_UNUSED(x) static_cast<void>(x)
+#define _FALLTHROUGH __fallthrough /* annotates depending on your compiler, sometimes does nothing */
+#else
+#define _MAYBE_UNUSED(x) static_cast<void>(x)
+#define _FALLTHROUGH /* no standard way to do it */
+#endif /* (__cplusplus >= 201704L) */
```
SyrianBallaS, line 30: This file will be added to periodically, but this only manages _language_ features since that is easy to mange.
Managing _library_ features is not recommended and should just use whatever libraries we are allowed to use.

src/ffscript.h
```diff
@@ -548,7 +548,7 @@ void do_savegamestructs(const bool v, const bool v2);
 void do_loadgamestructs(const bool v, const bool v2);
 
 int combo_script_engine(const bool preload);
-int FFScript::combo_script_engine_waitdraw(const bool preload);
+int combo_script_engine_waitdraw(const bool preload);
```
SyrianBallaS, line 551: Using scoped member function name in the declaration is not allowed by the standard.

src/link.h
```diff
@@ -518,12 +519,12 @@ void paymagiccost(int itemid, bool ignoreTimer = false);
...
-const int SEL_DOWN = 2;
-const int SEL_RIGHT = 3;
-const int SEL_VERIFY_LEFT = 4;
-const int SEL_VERIFY_RIGHT = 5;
+constexpr int SEL_UP = 0;
+constexpr int SEL_LEFT = 1;
+constexpr int SEL_DOWN = 2;
+constexpr int SEL_RIGHT = 3;
+constexpr int SEL_VERIFY_LEFT = 4;
+constexpr int SEL_VERIFY_RIGHT = 5;
```
SyrianBallaS, line 527: For C++11 and up it will use `constexpr`, pre-C++11 it will use `const`. See `compat_config.h`. There are many changes like this.

src/maps.cpp
```diff
@@ -641,6 +641,7 @@ int WARPCODE(int dmap,int scr,int dw)
 void update_combo_cycling()
 {
     int x,y;
+    _MAYBE_UNUSED(y);
```
SyrianBallaS, line 644: Normally, it would be done like this;
```
int x, y [[maybe_unused]]; // or
int x, y [[gnu::unused]]; // or
int x, y __attribute__((unused));
```
To satisify the requirements a macro was made. C++17 is checked first since that is where the standard attribute is defined, next it's based on compiler.

src/parser/Scope.cpp
```diff
@@ -74,7 +74,7 @@ Scope* ZScript::lookupScope(Scope const& scope, string const& name, bool noUsing
 		for(vector<NamespaceScope*>::iterator it = namespaces.begin();
 			it != namespaces.end(); ++it)
 		{
-			Scope* usingscope = *it;
+			const Scope* usingscope = *it;
```
SyrianBallaS, line 77: This is a good way to tell that `usingscope` is a non-owning reference, so this `usingscope` must not be deleted. Many other changes like this.

src/parser/Scope.cpp
```diff
@@ -148,8 +148,10 @@ RootScope* ZScript::getRoot(Scope const& scope)
 
 // Lookup
 
-DataType const* ZScript::lookupDataType(
-	Scope const& scope, string const& name, ASTExprIdentifier& host, CompileErrorHandler* errorHandler, bool isTypedefCheck, bool forceSkipUsing)
+const DataType* ZScript::lookupDataType(
+	const Scope& scope, const string& name,
+	const ASTExprIdentifier& host, CompileErrorHandler* errorHandler,
+	bool isTypedefCheck, bool forceSkipUsing)
```
SyrianBallaS, line 154: Passing by `reference to const` is much faster than just `reference` if it doesn't require modification.
Basically `reference to const` is for `in` values and `reference` is for `out` values.

When the time comes, passing by `rvalue` reference should be preferred for moving ownership for efficiency.

src/parser/Scope.cpp
```diff
@@ -413,7 +415,7 @@ inline void ZScript::trimBadFunctions(std::vector<Function*>& functions, std::ve
 
 		// Check parameter types.
 		bool parametersMatch = true;
-		for (int i = 0; i < parameterTypes.size(); ++i)
+		for (size_t i = 0; i < parameterTypes.size(); ++i)
```
SyrianBallaS, line 418: unsigned/signed mismatches in `for` loops usually lead to infinite loops.

src/qst.cpp
```diff
@@ -2603,6 +2603,7 @@ int readheader(PACKFILE *f, zquestheader *Header, bool keepdata, byte printmetad
 				{
 					switch(tempheader.build)
 					{
+                        case 0: _FALLTHROUGH;
```
SyrianBallaS, line 2606: This is to prevent `no case label` warning.

src/zc_sys.cpp
```diff
@@ -7180,13 +7182,14 @@ int onVidMode()
 //Added an extra statement, so that if the key is cleared to 0, the cleared
 //keybinding status need not be unique. -Z ( 1st April, 2019 )
 
-static enum uKey
+/*static*/ enum uKey
```
SyrianBallaS, line 7185: This `enum` is never used in this file so using `static` doesn't make sense.

src/zfix.h
```diff
@@ -6,19 +6,24 @@
 #ifndef ZFIX_H
 #define ZFIX_H
 #include "zdefs.h"
+#include "compat_config.h"
 #include <math.h>
 #include <limits>
+
+#ifndef NAN
 #define NAN 0
+#endif // !NAN
```
SyrianBallaS, line 15: `NAN` is defined in newer versions of Windows, if it's not defined we will defined it.

src/zq_tiles.cpp
```diff
@@ -16753,23 +16754,24 @@ static ComboAttributesInfo comboattrinfo[]=
 	},
 	{
 		cDAMAGE1,
-		{ "Custom",NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL},
-		{ "Amount",NULL,NULL,NULL},{ NULL,NULL,NULL, NULL}
+        // TODO: Placing these as globals can prevent the need of casting
+		{ const_cast<char*>("Custom"),NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL},
+		{ const_cast<char*>("Amount"),NULL,NULL,NULL},{ NULL,NULL,NULL, NULL}
```
SyrianBallaS, line 16759: In ISO C/C++ (starting with at least C99)
Casting away `const` is undefined behavior but for now, I just used `const_cast`.

SyrianBallaS (commented):
Left some review comments to explain some things. 
I noticed some formatting looks different here compared to IDE.

I'll try to fix those manually. We could benefit from using `.clang-format` to enforce a formatting standard.

src/parser/ffscript.ypp
```diff
@@ -284,7 +284,7 @@ Namespace :
 		ASTExprIdentifier* idens = (ASTExprIdentifier*)$2;
 		std::vector<std::string> components = idens->components;
 		std::string name = components.front();
-		int size = idens->components.size();
+		const int size = idens->components.size();
```
SyrianBallaS, line 287: Some stuff in this file was easy to fix. Most of it was hard to follow so was ignored, analyzer was showing 48 const rule violations in `ytab` but ignored them as they'd have to fixed in this file but was too afraid to mess this file up.

EmilyV99:
This has conflicts, and is probably quite outdated, especially now that "2.55-master" is on C++17. Closing w/o merge; if you feel like fixing merge conflicts and/or updating for C++17 standards, feel free to submit a new PR.

PR #654, opened 2020-10-02
Parser Memory Leak Fix

SyrianBallaS:
This PR's purpose is to fix memory leaks in the ZScript parser. Not all leaks are gone but most are. Will continue to try to fix the other memory leaks after this PR.

Changelog:
- VLD support added, requires that you install it as it's completely optional
   - Gives you a stack trace of the memory leak
   - It is windows only
   - Requires PCH and more instructions are in `precompiled.h`
- Boost was added to aid in removing memory leaks
   - Everything works right out of the box, no need for extra work other than configuring cmake how you want
- Allegro 4.4.4 was integrated
   - The files changed are the ones pre-built in ArmegeddonGames/allegro4


ZoriaRPG:
Does this require aught other than pull and build? Is everything set up tp build w/o extra steps? If not, what are the requisite steps needed to build?

SyrianBallaS:
There are no extra steps. Unless someone wants to debug memory leaks with VLD then they need to install it from the website (installer takes care of all the paths and VS hooks for VS 2008 and above). Boost stuff is already taken care of. Allegro dll/lib and zcsound lib are replaced.

PR #658, opened 2021-01-08
Fix hopping softlock

EmilyV99:


EmilyV99:
Does this need a QR? If not, then should be good to merge, and backport to 2.53 branch

EmilyV99:
Needs QR, backport for 2.53 needs emulation setting.

PR #678, opened 2021-03-18
Work/2.55 x y button items

ZoriaRPG:


Alucard648:
Will it be possible to add items that unlock X/Y item button usage when Link obtains them, just like in A Link Between Worlds?

PR #704, opened 2021-04-17
Fixed sideview vires and f6 mouse.

arceusplayer11:
Fixed Vires not working in sideview.

Fixed mouse not disappearing on f6.

### DIFF
--- a/src/guys.cpp
+++ b/src/guys.cpp
@@ -3257,7 +3257,7 @@ void enemy::move(zfix dx,zfix dy)
 	if(family >= eeSCRIPT01 && family <= eeFFRIENDLY10 ) return;
 	}
 	*/
-	if(!watch && (!(isSideViewGravity()) || isOnSideviewPlatform() || !enemycanfall(id)))
+	if(!watch && (!(isSideViewGravity()) || isOnSideviewPlatform() || !(moveflags & FLAG_OBEYS_GRAV) || !enemycanfall(id)))
 	{
 		x+=dx;
 		y+=dy;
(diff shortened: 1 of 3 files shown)


ZoriaRPG (commented):
if (tempguy.misc9!=e9tPOLSVOICE&&tempguy.misc9!=e9tVIRE) tempguy.moveflags = FLAG_OBEYS_GRAV | FLAG_CAN_PITFALL;

Did vires fall in 2.50.x? if not, then this is wrong.

ZoriaRPG (commented):
eat_buttons()

You are commenting out eating the map button. This on the surface appears likely to cause issues.

arceusplayer11:
Vires did not fall in 2.50; but I figured the reason they broke was Rob trying to allow them to fall in sideview. I kept his behavior (broken as it is), but it now requires the moveflag being checked, and it defaults to unchecked for vires in older quests.

Also, that was my error. I was attempting to fix inputs not working on the pause screen; it's a weird issue because it eats every keyboard input, such as enter and f6, while greyed out/paused, but not while the game is active, which leads me to believe it's an issue somewhere in the code. I was unable to resolve the error, but I haven't exactly gotten the hang of branching fixes and etc yet. Fixed now.

Also, that is not the map button; I was confused too! Map button is referred to as "P" instead of "M"; "M" is menu (which is defaulted to the escape key).

ZoriaRPG (commented):
I am not sold that vires should have the change as inserted, but I would approve this. The issue is that I am unsure if approving this would also pull any of your other changes in the npc-?>Data PR, and if they did, I cannot approve it, as that is hellishly broken.

arceusplayer11:
This is a separate branch.

PR #705, opened 2021-04-17
Added npc->Data.

arceusplayer11:
Added npc->Data as a script variable and as an editor box. This is a ghost exclusive variable intended for usage with ghost. It does nothing besides store a value; I don't see this being a problem. I know this probably goes against your standards Zoria, but my motivation for Ghost 3 hinges on this and if you let this slide I'll owe you one (as in I'll probably do something on your todo list for you)


ZoriaRPG (changes requested):
This should be an array sized [8] and named GhostData[].

Note: Writing to ZASM arrays does not have the access issues as ZScript POD arrays. It requires an additional PUSH and POP for INDEX, but no more, and in my experience, we will likely want to expand this. 

Note that while we can overload functions, we cannot overload variables. Thus, we can have foo(int), foo(int, int), but not int Data, and int Data[8].

I absolutely object to calling it ->Data, as this may be needed in the future. 

Last, you need to copy this to eweapon (and lweapon). 

I suspect that you want this for Combo draws, which as I said should only ever be done via the header, but if you want generic ghost3 data for these objects, then I do not have an issue adding them under that premise.

arceusplayer11:
Renamed npc->Data to npc->GhostData; also added GhostVX, GhostVY, GhostAX, GhostAY, GhostFlags, and GhostMisc[8].

Ghost Flags also has editor checkboxes to go with it; these will correspond to the flags ghost3 currently uses, so the names are placeholder for now but they work as intended.

ZoriaRPG (changes requested):
This is now even more broken. You cannot have a class array that is not fully initiaised in every instance of creating that class. The array requires a full initialiser in each type of instance, or its indices will point to random memory.

Further, the ZASM command is not treating it as an array. I also specified 8 indices, to match every other D[] type array in ZASM.

This needs to be treated as a size [8] array at all levels. At present, it is a memory hole. It is not properly initialised in any area, including guys.cpp, qst.cpp, and in ffscript.cpp it is not treated as an array (nor do I see any attempt in globalsymbols.cpp to do this).

Please look at how arrays are initialised in classes. Every **constructor** requires full array init, or you create memory fragmentation in their class instances and ppoint to invalid memory in every instance of the class!

I would prefer this be pushed as a branch so that it can be corrected, but your other changes use it as a foundation. It is a pretty awful mess right now.

arceusplayer11:
My other changes do *NOT* use this as a foundation. This is it's own branch.

EmilyV99:
This has conflicts now. Closing PR. New one can be submitted if you feel like fixing the merge conflicts, Dimi!

PR #706, opened 2021-04-19
Reorganized Combo Editor.

arceusplayer11:
Combo Editor now uses subtabs instead of the annoying scroll arrows.


ZoriaRPG:
The new tab should be on the right of Weapon triggers, and must specify (unimplemented).

I am also unsure if this is something that I want to spend time implementing, given that the existing feature has bugs. Best thing I suppose is to put it in and hide it. I still need to review the code to ensure that anything that you added to the combos structs isn't broken. I see new flags and other data here, and as the npc->Data commit shows me, you are still unfamiliar with c/c++ initialisers and constructors.

arceusplayer11:
The new tab isn't even in this commit. It's just the reorganization.

PR #720, opened 2021-07-05
Add GitHub actions CI

connorjclark:
This simply builds zquest and zelda targets on every PR, as a quick verification that changes at least build properly.

Only builds for windows. Toolchain is MSVC 19.29.30038.1

(to actually see it run, a repo admin needs to accept "allow actions from new contributors" near the bottom of this page. Here is it running on my fork: https://github.com/connorjclark/ZeldaClassic/runs/5299387281?check_suite_focus=true )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [2.55]
+    branches: [2.55-master]
   pull_request: # run on all PRs, not just PRs to a particular branch
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    - name: Install Dependencies
+      uses: crazy-max/ghaction-chocolatey@v1
+      with:
+        args: install .github/dependencies.config -y
+
+    - run: cp src/metadata/*.h.sig src/metadata/sigs/
+    - run: cmake . -A win32
+    - run: cmake --build . --config Debug --verbose --target zquest
+    - run: cmake --build . --config Debug --verbose --target zelda
```
connorjclark: wasn't sure if I should build Debug or Release here, any preferences?

EmilyV99: Release is more important to check for, sometimes debug will compile while release won't; never seen it the reverse.

connorjclark: (Sorry I went MIA!)

Changed it to the Release build.

PR #729, opened 2021-08-10
'Secrets->Item' flag, and 'Screen->SecretsTriggered()'

EmilyV99:


Alucard648:
What about also doing Secret->Shutters flag that opens standard NES Dungeon sutters, when secrets are triggered by any means?

PR #731, opened 2021-08-11
Item secrets

EmilyV99:


Alucard648:
What about Secret->Sutters, so any trigger or scripted puzzle can open old school NES Dungeon shutters and keep them open, when triggered/solved?

PR #739, opened 2021-09-09
Integrate TeZSLa Compiler Alpha v1.1.5

ZoriaRPG:
**WARNING** This reverts a minor feature  (font shadow colour)--that you can fix--to add a major feature:::

TeZSa Compiler Supports the following args:

-i inputfiename : The buffer/main script. ZQE sets this in the system() call.
-o outputfilename: Theintended ZASM outpt file. ZQE sets this in the system() call.
-r ruesfiename: Predefined rules for the compiler. ZQE sets this in the system() call.
-s settingsfilename: A settings file. Behavior undefined.
-logging int: Flag to set logging type. At present this and all settings (below) write to an array that is unused.
-division: int Flag to set if we use old div bug.
-short int: Flag to set short circuit on or orr (shod default to on in future)
-truedec int: Flag to enable/disable true decimal booeans.
-hguard int: Options for header guard.
-halt int: Options for error halting.
-trueint int: Fag for using true int size. Should be defaulted to on, in the future
-inline int: Flag to enable unimplemented inlining option.
-32b int: Flag to enable true 32b.
-margin float: Value of Appx Equal operation.
-caseins float: Stringcase insensitivity.

In the future we can change any of these that only have 0/1 option fro -switch int to -switch. Right now yo must specif values for each, as I did not yet add a defaults section. In an event these do not yet do anything as the array used isn't read by the parser itself, and must be supplied to it and coned to zquest.cpp so that its use always compiles with logical OR instructions.

What This Needs
1. Please please, please, learn how CPP linkers work across project applications. I wasted many hours fixing the src to remove linker issues.
	RAMPal cannot exist in that domain and the colour select for jwin needs to be segregated! 
	The way that you implemented shadow colour is not possible as-is: I disabled that to add this which is far more critical.
	Shadow colour can be fixed (e.g. misc colours) or you can work out a way to do it outside of jwin.cpp and RAMPal 
	or you can re-implement it and fix linker issues: IDK which.
	You use this in one spot and a palette selection check/UI with some vars--local or gobal--would do the smae thing w/o compromising the code across applications
2. Dump ZASM to an output file, and add the ability to read that dump.
3. Stripping all vars, function calls,and menu options not used by TeZSLa.
4. A custom DAT with a '.qst' file that ony holds temp script data.
5. Replace loading default or recent .qst with dummy script quest.
6. New menus and a proper Abort if invoked by CLI.
7. A shell in place of the box_out.
8. Integration with SciTE.


CMakeLists.txt
```diff
@@ -180,21 +184,21 @@ target_link_libraries(zcsound ${ALLEGROLIB} ${SOUNDLIBS} ${ZCSOUNDLIBSEXTRA})
 # Romview
 #############################################################
 
-# if(MSVC)
```
EmilyV99, line 183: Any particular reason you un-commented the romview section?

src/jwin.cpp
```diff
@@ -72,7 +72,7 @@ int scheme[jcMAX] =
 };
 
 int jwin_pal[jcMAX] = {0};
-extern PALETTE RAMpal;
```
EmilyV99, line 75: Having access to read the colors of the palette is needed for new features. Should you need jwin, you should have a palette. This was very intentionally added, and I would love to know why you felt the need to remove this.

src/util.cpp
```diff
@@ -200,71 +201,7 @@ namespace util
 #endif
 	}
 	
-	long long zc_atoi64(const char *str)
```
EmilyV99, line 203: Why was any of this removed??

src/zquest.cpp
```diff
@@ -1967,12 +2029,28 @@ static MENU media_menu[] =
...
+	return ret;
+	//return system("tezsla.exe -i buffer.zs -r tezsla.rules"); //need specific fienaes here which must be provided
+	//buffer file must be dumped and specified ese crash-poo --Z
+}
 
 //New ZScript Menu for 2.55 Alpha 16
 static MENU zscript_menu[] =
 {
 	{ (char *)"Compile &ZScript...",            onCompileScript,           NULL,                     0,            NULL   },
+	{ (char *)"Launch &TeZSLa...",              onTeZSLa,           NULL,                     0,            NULL   },
```
EmilyV99, line 2053: For this to be implemented, it should be *invisible* to the user; i.e. there should be no new menu option, it should be launched when they click "Compile" in the normal compilation dialog.

src/zsys.cpp
```diff
@@ -2396,6 +2424,73 @@ void safe_rect(BITMAP *bmp, int x1, int y1, int x2, int y2, int color)
     rect(bmp, vbound(x1, 0, bmp->w-1), vbound(y1, 0, bmp->h-1), vbound(x2, 0, bmp->w-1), vbound(y2, 0, bmp->h-1), color);
 }
 
+long long zc_atoi64(const char *str)
```
EmilyV99, line 2427: Why were these shoved here??

EmilyV99 (changes requested):
Why does "tezsla.cpp" have so many things from "zquest.cpp" that it doesn't need to use?
Please address comments.

CMakeLists.txt
ZoriaRPG: Still stripping those components but it will take a while. It needs to function in many ways like using a dat that contains only script sot data (empty by default) so that I can remove all of the calls that point to or expect a real quest in memory; without that it'd instantly segfault. Others can contribute to accomplishing that,too. The main thing now is to integrate it so that it does not lose parity with 2.55 as a while to allow easily integration and removal of boat.

The core features work and I restored the font shadow stuff in PR #740 implementing it where it used, rather than globally exposing it and requiring RAMPal to be a part of JWin. The ZASM/Slot dumps need further  work and integration, but the already respect -inputfie and generate some useful output. I think that my dump timing is off and that do_slots needs reworking.

TeZSLa is a work in progress to be sure, but it is a WIP that needs constant and regular parity with the base as  to not need to be regularly rewritten. I provided something that any of ou could begin to whittle down, although my next task is making the needed custom tes.dat.

ZoriaRPG: No. It was not removed on my end when I made this over a month ago, and I merely copied over the cmake files from m local repo. Why did you remove it?

EmilyV99: It isn't used by anyone for a along time, and adds needless compile time. It was commented out in the cmake files such that build time would not be wasted on something no one uses.

EmilyV99: > The core features work and I restored the font shadow stuff in PR #740 implementing it where it used, rather than globally exposing it and requiring RAMPal to be a part of JWin. The ZASM/Slot dumps need further work and integration, but the already respect -inputfie and generate some useful output. I think that my dump timing is off and that do_slots needs reworking.

Why is jwin having access to the palette an issue whatsoever?

src/jwin.cpp
ZoriaRPG: That implementation causes liner duplication conflicts with the added binary. I moved the new feature code to the only place that it is used or likely to be used it is static and if needed elsewhere can be thusly implemented. It still works as you intended without confusing the linker.

You should poll a three forms about romview, although the packaged version is not from this branch. i planned to fix the video issues with this version in 3.0 when I universal implement OpenGL.

EmilyV99: Why did that implementation cause any conflicts? That sounds like an issue with *your additions*, not ours.

src/zsys.cpp
ZoriaRPG: Same reason: Linker conflicts. Why move them at all? Honesty, go move the stuff back and see for yourself.  If you can compile with things as you set them up, fine. This change that has no useful purpose made lining to the new binary impossible. What is the value or motive of moving them out of zsys?

The entire point of zsys is for these commonly used functions.

EmilyV99: What the hell did you do to get linker errors from that?

EmilyV99: Util.cpp is for pure utility functions,  zsys has plenty of other crap in it.

src/zquest.cpp
ZoriaRPG: > For this to be implemented, it should be _invisible_ to the user; i.e. there should be no new menu option, it should be launched when they click "Compile" in the normal compilation dialog.

It's temporary; and Linux may or may not be immediately able to use it even when it is ready. Adding a menu to invoke the eternal versus internal compiler at this time is needed until both the external compiler is finished and until it is proven with Lx. It certainly needs a way to invoke and to test it easily, despite not yet being mainline.

In the end the goal isn't to make it invisible but to make it clear that it is external;  and to only use the external programme. Adding it as a menu issue is needed to test it but it s not a singular end goal.

Given how many partially complete features I allowed you to add that were far more incomplete and untested with no  user choice I see this argument as insane. Many of these you have yet to resolve but the user has no choice but to use them This is a choice and a way to test the programme without being locked or forced into it.You can always make a definition and make it so that until ready, pubic releases do not see it. I can add that as a compiletime definition it it solves your concern about user confusion.

PR #740, opened 2021-09-10
Restore font shadow colour dialogue.

ZoriaRPG:
Moved the code for msg string font coour from global jwin scope, to static invocations in zq_str, which should restore this feat, without linker issues.

What remains:

1. Finish stripping ZQ resources from TeZSLa
2. Correct and implement switches.
3. Add DAT file to use in place of qst.dat and always load that.
4. Finish ZScipt Packfiile encode and unpack hen TeZSLa exits.
5. Integrate SciTE.


EmilyV99:
1. Why does it need a quest file at all? Anything that isn't parser should be stripped away.
2. Why did you need to modify *ANY* of the jwin code at all?
3. Are you going to reply to any of the fixit comments on the previous PR?

ZoriaRPG:
Once compiled ZASM is assigned to slots. The output ofTeZSLa is still intended to be a packfile of the *slot data*. It needs an empty .dat to use to preload blank slots in packfile format.

ZoriaRPG:
P.S. if you have an outline and plan docs for another way to so this you now where to send them.

EmilyV99:
It shouldn't need a .dat to be able to output a packfile of slot data

ZoriaRPG:
> It shouldn't need a .dat to be able to output a packfile of slot data

You need a basic bank packfile of some sort with a new .dat with only the packfile data needed for allegro to assign slot data is required: Tell me how to get around that as the application needs a basic memory map to use *on init*.

EmilyV99:
Why would you need a .dat to store blank data? Just initialize the memory to blank,  no need to have that in a file...

ZoriaRPG:
I could explain why I did what II did to fix this into oblivion You either believe me or not but the issue for externing RAMPal will always exist It should never exist in that subframe as not only does that require RAMPal to exist as an object in anything that uses jwin but  any two programmes with shared resources in a project that  are linked that both use win then the compiler will fail to resolve what to use.
 
This also applies to adding functions to bound to as namespace the invoking the using namespace directive The compiler will have a resolution issue. This is why I advise reading The C Standard and The C++ Standard boos to those who study under me.
 
The things that I changed specifically fix these issues The fact that you want to overlook them s sheer arrogance IMO, as not only do my changes work and break naugh,  but problematically they make sense, and ensure future work
 
I am not reviling you for them, but they need discussion and actual debate. If you disbelieve me  then ask Samer.
 
I just want everything to function and not to breaking that need not be broken.
 
I will continue to work on what I feel that I must finish: If I must do that alone so be it but I would prefer to cooperate
 
If that is not possible then I will work on the compiler as a separate project that is not GPLv3 and you an elect to use  it or not.
 
You asked me to explain, so I did.

ZoriaRPG:
If you have specific changes that you require, please list them. if not then can we at least branch this? It is weeks at most from, completion.

EmilyV99:
> I could explain why I did what II did to fix this into oblivion You either believe me or not but the issue for externing RAMPal will always exist It should never exist in that subframe as not only does that require RAMPal to exist as an object in anything that uses jwin but any two programmes with shared resources in a project that are linked that both use win then the compiler will fail to resolve what to use.

So any binary that uses JWin must have a RAMpal. Why is that unreasonable?
And how would "any two programmes with shared resources in a project that are liked that both use win then the compiler will fail to resolve what to use"? In what way would the compiler fail? RAMpal should be declared *exactly once* per binary. If it is declared multiple times in a single binary, that would be an error in many files, not just jwin.
To be clear, this is not be just being an ass about something; I legitimately do not understand the reason you've done this.

> The things that I changed specifically fix these issues The fact that you want to overlook them s sheer arrogance IMO, as not only do my changes work and break naugh, but problematically they make sense, and ensure future work

From where I sit it looks like you meaninglessly reverting changes I made to make things more difficult for me. If I do not get a satisfactory and well-enough explained reason behind a change, I will deny any pull request containing that change. If it is not required for teszla to compile, simply create a PR that does not include it.

> I just want everything to function and not to breaking that need not be broken.

I still don't understand how *anything* was broken.

> If you have specific changes that you require, please list them. if not then can we at least branch this? It is weeks at most from, completion.

I require a well-enough explained reason for any change in a PR. Lacking that for any change in the PR, and the PR will be denied. I gave numerous comments on things asking reasons, and have yet to be satisfied with any of your explainations.

PR #761, opened 2021-11-17
2.55 emptybottles

EmilyV99:
BOTTLEEEEESSSSSSSSSSS!!!!! WOOHOO!
Still need to add script access for "Bottle Types" and "Bottle Shop Types", so scripts can modify their values; but that requires 2 new data types, which is a lot of work, so merging this now with plans to add that later.


Alucard648:
Will it be possible to have more than 4 bottles in a quest?

EmilyV99:
> Will it be possible to have more than 4 bottles in a quest?

As implemented, you can have up to 64 "bottle types" (things that go IN the bottles, 64 does not count the 'empty' state), and up to 256 empty bottles. Given, the item limit would become a problem for you if you tried to use them all, but scripts could use the bottle states for fancy scripty things, and it only costs 1 byte of memory per empty bottle allowed, so no reason NOT to allow a full 256.

Alucard648:
Thanks. Most quests, maybe, have 1-6 empty bottles and up to 32 possible content types.

PR #774, opened 2022-03-22
Upgrade to allegro 5

connorjclark:
This PR replaces allegro 4 with allegro 5, without actually modifying any of the a4 api calls.

You can download a build here: [release](https://drive.google.com/file/d/1J9RYIfILsqpLF6uFy_L-bIVLkjxpPLJb/view?usp=sharing)

The "Files Changed" tab is pretty rough because of the number of deletions and renames, and unfortunately GH is really laggy with so many changes. It's only the second commit that is super large (the one that replaces allegro for allegro-legacy folder), so a good option is to just view the diffs in two parts:

- [commit 1](https://github.com/ArmageddonGames/ZeldaClassic/pull/774/commits/c9eb3af59314bc5d4a75ae1e95bdcd63c0ca1c72) (adding allegro 5 / an example program)
- [commits 3+](https://github.com/ArmageddonGames/ZeldaClassic/pull/774/files/74589c175cb18ad38904b05b7e6734cca68348a9..7f71b55cb6595fccc39ad2314ae736bd6b099fa5) (the bulk of the changes)
______

Benefits of this PR:
- Upgrade from 4.2.2 (circa 2007) to 5.2.7.0 (Mar 2021)
- Removes a barrier for eventual 64-bit targets (would still need to replace the 32bit allegro addons, among other things)
- Allegro 5 supports using SDL as the system backend, which has enabled me to port Zelda Classic to the web via emscripten (actively working on this)

To do that, it uses [Allegro Legacy](https://github.com/NewCreature/Allegro-Legacy), which was nearly a drop-in replacement (I discovered a few bugs which had to be [fixed](https://github.com/NewCreature/Allegro-Legacy/pulls?q=is%3Apr+is%3Aclosed)) 
Allegro-Legacy emulates the a4 api layer with calls to a5, and seems to do it pretty well. Thanks @NewCreature!

Because Allegro-Legacy provides all the symbols that a4 did, the pre-built a4 addon libraries will continue to just work.

The Allegro-Legacy project is committed to source, and replaces the allegro folder (although you'll see a bunch of renames in the source history because allegro-legacy copied the a4 project).

The files in `allegro-legacy/src/a5` are the "glue layer" between a4 and a5. In that folder, you will find code that start new threads for particular concerns, each with their own event queue. `a5_system.c` is what glues it all together: it creates an a4 System driver which implements the old a4 interfaces using new a5 stuff.

Allegro5 is not commited directly to source, instead it is built from source as a CMake dependency.

I also included a simple allegro 5 example project for purposes of testing, perhaps it'll be useful for simply confirming your local dev environment works on a basic level.

What this PR does NOT do:
- rendering on the GPU. all draw calls are still happening on the CPU. GPU/shaders are now an option, though it'd take a good amount of work to take proper advantage of this
- explictly use any of the new a5 event queues. Implicity, allegro-legacy has multiple threads with an event queue to interface with the allegro 5 system (see allegro-legacy/src/a5)

Some misc. stuff:
- Added "-debug" postfix to debug lib targets, to help distinguish build artifacts
- Set the VS_DEBUGGER_WORKING_DIRECTORY for VS Code, which allowed me to use the debugger in the IDE
- There's a "build-emscripten.sh" file that does nothing more than build the allegro 5 example project for the web. When I initially wrote this PR, at the same time I was curious
  and wanted to verify that I was setting things up to make compiling with emscripten possible

From what I could tell, there is no impact on performace.


connorjclark:
> Added "-debug" postfix to debug targets, to help distinguish build artifacts

I've realized this breaks the zlauncher for debug builds. that's... probably fine? I could revert this change or pretty easily patch zlauncher, lmk.

fm3845w0:
Very cool! The Allegro Legacy API emulation as a workaround for update difficulties is some ingenious lateral thinking and a great start to a ZC project so ambitious it was long called impossible. I hope you succeed!

How much easier would your work make porting ZC to a libretro core for use in RetroArch and other programs?

PR #778, opened 2022-06-11
remove snprintf alias in msvc, and add minimum compiler check

connorjclark:
Since ~2010 MSVC has supported `snprintf` so no need to alias it as `_snprintf`, which is _slightly_ different in two ways:

1) return value. Doesn't matter, we never use the return value
2) edge cases re: insertion of \0 null byte. Don't need to hardcode the null byte into the format string to avoid this edge case, so can drop all instances of \0 (which will help with enabling the -Wformat compiler warning as an error)

See [this ](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170#remarks) for more.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,11 @@ endif()
 if(WIN32)
 	# Prevents definition of min and max macros, which conflict with std::min and std::max
 	add_definitions(-DNOMINMAX)
+
+	string(REGEX MATCH "MSVC" CMAKE_COMPILER_IS_MSVC "${CMAKE_C_COMPILER_ID}")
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -116,6 +116,11 @@ endif()
 if(WIN32)
 	# Prevents definition of min and max macros, which conflict with std::min and std::max
 	add_definitions(-DNOMINMAX)
+
```
connorjclark, line 119: https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd

PR #817, opened 2022-11-15
2.55 paldata2

Moosh64:


connorjclark:
If it'd make sense to squash this to one commit - what should the commit message be?

And could you write a bit about the feature - sorry, I don't understand palette stuff too well, but maybe some example scripts would help me understand. Is this just a scripting feature?

EDIT

🤣 

> Sorry I missed your comment on github. Emily was just walking me through the pull request process. Paldata's just a scripting feature and shouldn't interfere with anything you're working on.

PR #820, opened 2022-11-22
Solid objects

EmilyV99:


ZoriaRPG:
This is very useful , thank you. Does it include npcs, or just ffcs to
date? Solid npcs, including the ability to allow them push you or you to
push them (independent), is sort of. JRPG staple and would allow making
very good towns.

I did a tiny bit on solid ffcs years ago and was sickened by how all of
Link's hitbox stuff was so spread out and a mess.

If you ever feel up to it, one thing I would advise is to consolidate all
of the hitbox checks for Link to something sane. This includes interaction
with combos, enemies, weapons, and a bunch of other stuff.

Mostly this is important for 1x2 Link, such as Zelda 2, but it also applies
for z3 sized sprites, and solidity would equally be a factor there.

ZoriaRPG:
WRT solid damage combos, I had considered making them work more like
pushblocks: Use a clk, locking in the dir on say, frame 1, and on frame 2
damaging after pushing on them, and applying proper knockback via a
read-ahead.

Essentially finding a way to prevent damage clipping through them.

Alucard648:
Missing crushing between solid objects and combos.

EmilyV99:
...yes, we are very well aware of this. Almost as though it's on our todo
list.

connorjclark:
> Missing crushing between solid objects and combos.

BTW @Alucard648 PR comments aren't the forum to make feature requests (esp. not a month after it already merged?)

PR #822, opened 2022-11-23
Bump FFC Count (32 -> 128)

EmilyV99:


ZoriaRPG:
Did you do a RAM calc for this ? Ffcs are part of mapscr, so this is pure
bloat for absolutely no reason. When do you need > 32 ffcs?

Now that both sprites and combos can run scripts, ffc use should be
downsized, not intentionally upsized. WTF is the goal of this? If you Calc
mem usage you will revert this ere anyone uses it.

Literally every dev before and parallel to me avoided exactly this because
of the long term ramifications. This singular decision is dooming for the
entire basecode in the future.

arceusplayer11:
ok boomer

connorjclark:
mapscr's ffcdata will be changed to a vector later, so it will only take up as much memory as called for. Not everything needs to be a fixed size array. Your comment isn't very helpful, why are you so mad?

ZoriaRPG:
Not mad, concerned. I have very little actual emotion about much of
anything, and anger is too strong of an emotion to waste energy exerting.

DO NOTE that you need to update the SAVESRAM function that handles mapdata
in ffscript, so that it also stores a vector size and vector elements. As
is, saving macspr via script would be wrong, and loading it would (well,
99.999% would) **crash**. As long as it is dynamic, it is fine, but you
need to convert those commands to handle it.

In general, if you make it dynamic, mat as well use a general ffc count,
starting at 0. You can make 32 the default, but allow people to reduce
that. This would allow people to electives limit this use and make both
quest size and RAM use smaller.

(Do ensure to manually check in debug, yhat the vectors are being killed
off the heap though. )

HIf it were static, having unused an extra 96 ffcs per mapscr would be
disgustingly wasteful. (ffc count struct size × NUMMAPSCR × mapcount).
Gleeok and I had previously debated making it a dynamically sized amount,
but never got around to it.

In general, mapscr is the largest RAM object in the engine, and as it is in
constant use per above formula. As such, **increasing ffc count was one
thing that the prior devs expressly forbade me to do**.

I didn't even want to add stuff like arg strings as the RAM usage and .qst
file size would quickly snowball.

Just don't forget to update the SRAM stuff in the process. You may need to
look into whether other new additions to constant memory objects that
Saving and Reading SRAM (Script RAM), aren't doing, or are doing improperly.

Have fun and good luck.

ZoriaRPG:
One last thought: once it is a vector, Screen-> and mapdata-> CreateFFC()
would be possible and useful. (I am not sure that DestroyFFC() would ever
be safe, but expanding the vector during gameplay would be viable.)

PR #823, opened 2022-11-23
use ccache in ci for faster builds

connorjclark:
For windows, now using ninja instead of msbuild, since ccache requires ninja. Just comparing ninja vs msbuild with all cache misses, looks like ~30% faster.

For all cache hits, the build step went from ~15m to ~10s (mac) and ~40s (windows).


connorjclark:
and the web build step went from 42m to 19m

PR #843, opened 2022-12-21
Use MessageBox on Win32 to display fatal errors

leximayfield:
Whenever a fatal error happened, Zelda Classic used to simply die with no message on Windows.  Now it displays a message box with the error message if the zquest console isn't up.

image

This is my first PR for this project, please let me know if I overlooked something.  In particular, due to the mix of spaces and tabs I went with whatever the editor spat out by default.


connorjclark:
Thanks!

Allegro has `al_show_native_message_box` – would that work here too, so we can support all platforms? see https://www.allegro.cc/manual/5/al_show_native_message_box

connorjclark:
we didn't intend to merge–but if you'd like to followup using allegro's apis that'd be appreciated!

PR #848, opened 2023-02-24
A5 render tree upgrades + CSets 12 & 13

EmilyV99:


src/base/render.cpp
```diff
@@ -46,7 +58,22 @@ static void render_tree_draw_item(RenderTreeItem* rti)
 			al_draw_scaled_bitmap(rti->bitmap, 0, 0, w, h, rti->computed.x, rti->computed.y, w*rti->computed.scale, h*rti->computed.scale, 0);
 		}
 	}
-
+	
+	for(ALLEGRO_BITMAP* overlay : rti->overlays)
```
connorjclark, line 62: It's strange to call this an overlay when the node's children will be drawn over this. Can you think of a better name?

Seems like it'd be equivalent to put wrap these A5 bitmaps in a RenderTreeItem (with a null a4 bitmap) and make them the first children of this node. Does `rti_dialogs` even use its children? If not, should be simple to change the add/remove overlay stuff to add/remove RTIs from `rti_dialogs` children.

EmilyV99: noted, merging anyway. Can clean up later.

PR #881, opened 2023-08-20
fix: GCC 13 Compilation Errors

frutiemax92:


connorjclark:
I pushed a few commits to fix gcc.

Only remaining issue is with GME, will fix it upstream. https://bitbucket.org/mpyne/game-music-emu/issues/242/issues-compiling-with-gcc

PR #889, opened 2023-10-30
Cache 'item_tile_mod'

EmilyV99:


connorjclark (approved):
This is going to be really hard not to break. It's not clear offhand what values would necessitate clearing the cache for `item_tile_mod`. What would help w/ that is at least a comment of each property that loudly says "you must call `flushTilemodCache` if you modify this. But then that just begs for a better interface for modifying these values.

For the few values in Hero that would invalidate this cache: it'd be cleaner to maintain a separate copy of each value in a `item_tile_mod_state` struct or whatever, and compare against every frame to check if work got invalidated. That would remove any potential future errors where modification of those Hero variables forgot to reset this cache, without introducing an annoying `set_bunny` or whatever that all code would be forced to use. Can't do the same with `itembuf[].ltm` because there are too many.

It seems the largest cost of `item_tile_mod` is iterating over everything in itembuf simply to  load a single bit `ITEM_BUNNY_ENABLED` is not set. In the common case there are next-to-no bunny items yet we load all this stuff into memory. Maybe we need to cache a `std::vector<id> bunny_items` instead? EDIT: Emily pointed out this only is the case when bunny mode is active

connorjclark:
Working on this in another commit. Thanks for the first pass.

PR #890, opened 2023-10-30
Spotlight caching

EmilyV99:
Refactored the `cpos_info` code, now can cleanly cache any number of bits of information about the combos on-screen. This PR uses this to cache whether or not any cSPOTLIGHT combos are present, and skips portions of `handleSpotlights()` when 0 such combos are present.

### DIFF
--- a/src/base/cpos_info.cpp
+++ b/src/base/cpos_info.cpp
@@ -2,6 +2,8 @@
 #include "base/zdefs.h"
 #include "base/combo.h"
 
+void update_cpos_cache(word oldid, word newid);
+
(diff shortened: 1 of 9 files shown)


connorjclark (approved):
Looks good.

Only thing I might suggest changing would be the function names. It's nice to prefix functions with common behavior w/ a common prefix. That's typically something you use namespaces for but I find a prefix (like `ttip_`, `render_`, `zapp_`, etc...) does the job just as well. So maybe `cache_` or just `cpos_` for these cache related queries?

PR #924, opened 2024-04-27
Array params

EmilyV99:


tests/scripts/errors/errors_3.zs
```diff
@@ -14,15 +14,33 @@ global script Global
...
 		Object objects3[] = {obj};
-		auto objects4[] = objects3; // TODO: this should be valid.
-		auto numbers3[] = {1, 2, 3}; // TODO: this should be valid.
-
+		auto objects4[] = objects3;
+		auto numbers4[] = {1, 2, 3};
+		untyped mixed1[] = {1, obj, 3};
+		auto numbers5[10] = {4, 5, 6};
+		auto enemies1 = Screen->NPCs; //is 'const npc[]'
+		auto enemies2[] = {enemies1[0], enemies1[2], enemies1[4]}; //is 'npc[]', despite 'const npc[]' elements
```
connorjclark, line 28: do const arrays in zscript only prevent modification of the array (setting an index or changing its length)?

tests/scripts/errors/errors_4.zs
```diff
@@ -34,13 +34,4 @@ global script Global
 		max_2 = Max(1, 2, 3);
 		max_2 = Max(1, 2, 3, Max(3, 4));
     }
-
-	// The template checking code explicitly allows char32 as an "array" type,
-	// for compat.
-	void old_ptrs(char32 ptr)
```
connorjclark, line 40: can we keep this to test when the `OLD_ARRAY_TYPECASTING` option is on?

tests/update_snapshots.py
```diff
@@ -6,15 +6,14 @@
 
 script_dir = Path(os.path.dirname(os.path.realpath(__file__)))
 
-
```
connorjclark, line 9: I need to get you configured with the python auto formatter I used on the code base and have set up on save in vscode. (it's called Black).

connorjclark (approved):

tests/scripts/errors/errors_3.zs
EmilyV99: Constness prevents assignment only. The new array typing DOES make this pass through the index operator, though.

So, `enemies1 = NULL;` will fail (can't change the pointer) and `enemies1[0] = NULL;` will ALSO fail (can't change the value)

In this case, the internal array from `Screen->NPCs` would issue a runtime error if you tried to write it's index (the second case); and the first case is how it has worked historically. Having it work like this... kinda happened just as a consequence of how the system works, and it seemed like an ok/simple enough way to leave it.

Function calls do *not* handle const qualifications though, so passing a `const int[]` to `ResizeArray()` is completely allowed. (In the case of `Screen->NPCs` here, this would be a runtime error, similar to how in C++ it would be undefined behavior to cast away constness of an array that was *declared* const - though it would work fine on normal user arrays, similar to casting away constness of a pointer to a mutably-declared array in c++)

PR #933, opened 2024-06-14
feat: add enemy flag to control if affected by "Kill All Enemies"

ghost:
Ringleader Rooms, Cheats, and KillAllEnemies Items will now kill adult Digdoggers.

### DIFF
--- a/src/qst.cpp
+++ b/src/qst.cpp
@@ -9796,6 +9796,11 @@ void init_guys(int32_t guyversion)
                     guysbuf[i].flags2 |= guy_transparent;
                 }
             }
+
+			if (i == eDIG1 || i == eDIG3)
+			{
+				guysbuf[i].flags2 |= guy_ignorekill;
+			}
         }
         
         // Darknut fix
@@ -15014,6 +15019,14 @@ int32_t readguys(PACKFILE *f, zquestheader *Header)
 				if(tempguy.family == eeWALK && tempguy.misc9 == e9tPOLSVOICE)
 				{
 					tempguy.moveflags |= FLAG_CAN_WATERWALK;
+				}
+			}
+
+			if (guyversion < 47)
+			{
+				if (tempguy.family == eeDIG && tempguy.misc10!=1)
+				{
+					tempguy.flags2 |= guy_ignorekill;
 				}
 			}
 			
--- a/src/zc/guys.cpp
+++ b/src/zc/guys.cpp
@@ -1743,7 +1743,7 @@ void enemy::leave_item()
 // auomatically kill off enemy (for rooms with ringleaders)
 void enemy::kickbucket()
 {
-	if(!superman)
+	if(!superman && !(flags2&guy_ignorekill))
 		hp=-1000;                                               // don't call death_sfx()
 }
 
@@ -13546,8 +13546,6 @@ void eLilDig::draw(BITMAP *dest)
 
 eBigDig::eBigDig(zfix X,zfix Y,int32_t Id,int32_t Clk) : enemy(X,Y,Id,Clk)
 {
-	superman=1;
-	
 	SIZEflags = d->SIZEflags;
 	if ( ((SIZEflags&guyflagOVERRIDE_TILE_WIDTH) != 0) && txsz > 0 ) { txsz = d->txsz; if ( txsz > 1 ) extend = 3; } //! Don;t forget to set extend if the tilesize is > 1. 
 	//al_trace("->txsz:%i\n", txsz); Verified that this is setting the value. -Z
--- a/src/base/zdefs.h
+++ b/src/base/zdefs.h
@@ -150,7 +150,7 @@ enum {ENC_METHOD_192B104=0, ENC_METHOD_192B105, ENC_METHOD_192B185, ENC_METHOD_2
 #define V_ICONS            10 //Game Icons
 #define V_GRAPHICSPACK     1
 #define V_INITDATA        39
-#define V_GUYS            47
+#define V_GUYS            48
 #define V_MIDIS            4
 #define V_CHEATS           1
 #define V_SAVEGAME        41
@@ -1362,6 +1362,7 @@ struct item_drop_object
 #define guy_transparent 0x00008000
 
 #define guy_ignoretmpnr 0x00010000
+#define guy_ignorekill  0x00020000
 
 // Old flags
 #define weak_arrow		0x20000000
--- a/src/zq/zq_custom.cpp
+++ b/src/zq/zq_custom.cpp
@@ -1589,12 +1589,12 @@ static int32_t enedata_data4_list[] =
 
 static int32_t enedata_flags_list[] =
 {
-	74,75,76,77,78,79,80,81,82,83,84,85,86,87,88,89,98,99,100,101,102,103,104,105,106,120,121,-1
+	74,75,76,77,78,79,80,81,82,83,84,85,86,87,88,89,98,99,100,101,102,103,104,105,106,120,121,400,-1
 };
 
 static int32_t enedata_flags2_list[] =
 {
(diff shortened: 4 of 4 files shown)


EmilyV99:
This change would break any old quest that relies on... well, this NOT happening. This would need a compat QR added.

connorjclark (changes requested):
this was added here https://github.com/ZQuestClassic/ZQuestClassic/commit/39369a5f6f4c5822ddc274af7137c7b64092d8e1

other enemies classes always have `superman` set too:

moldorm
bigdig
lanmola
manhandla
gleeok

(may be more)

some others I didn't list set it because it seems like you really would never want them to get deleted from the cheat "kill all enemies". like the item fairy.

any guess why this was set in the first place?

> This change would break any old quest that relies on... well, this NOT happening. This would need a compat QR added.

I think we need to articulate a real way this would break a quest before jumping to a QR.  Our replay tests are becoming pretty comprehensive, and nothing failed there. Every new QR we add is a burden into the future that can't be removed. We can always add a QR later.

but - if a QR is to be added, these other enemy types should be included in it.

src/zc/guys.cpp
```diff
@@ -13546,7 +13546,7 @@ void eLilDig::draw(BITMAP *dest)
 
 eBigDig::eBigDig(zfix X,zfix Y,int32_t Id,int32_t Clk) : enemy(X,Y,Id,Clk)
 {
-	superman=1;
+	//superman=1; !This stops the adults from dying when KickBucket() is called! -Jambu
```
connorjclark: please just delete lines of code

And you dont need to sign stuff, `git blame` lets you know who wrote what

connorjclark:
`kickbucket` effects:

- the "kill em all" item type
- the "kill" cheat (which oddly also has its own mechanism for ignoring enemies, `guy_doesntcount`)
- the "kill em all" combo trigger
- leaders dying triggering killing the rest of the enemies
- anything else?

but does not impact scripting npc Remove (despite the ZASM being called NPCKICKBUCKET...)

Mostly related to cheats or cheat-like weapons. I can't see a quest relying on this behavior, so I suggest we don't need a QR compat rule.

EmilyV99:
> I think we need to articulate a real way this would break a quest before jumping to a QR.

Quests may expect their boss not to die to an instakill ability that can possibly drop from the boss's minions in a dropset.

connorjclark:
would a new enemy flag for "ignored by kill all enemies" be better than a qr?

EmilyV99:
> would a new enemy flag for "ignored by kill all enemies" be better than a qr?

Aye, that sounds perfect

ghost:
A Behavior Flag it is then?

connorjclark:
Yes, let's do that. I think we can also consider making the "Kill all enemies" _cheat_ ignore the flag too, but that can come later.

I think the next flag would be 0x00020000 on `flags2`? https://github.com/ZQuestClassic/ZQuestClassic/blob/48660bfcbfad6ef6890913e9e565945e6deb3f4a/src/base/zdefs.h#L1364C1-L1364C35

Wow, we really haven't added a new flag in over 8 years. I'm not certain what the process is, but it will involve adding code in `readguys` / `writeguys`; and `readguys` will need to check the `guyversion` to set the existing enemies that are excluded from "kill em all" but won't be in new versions. [for example](https://github.com/ZQuestClassic/ZQuestClassic/blob/48660bfcbfad6ef6890913e9e565945e6deb3f4a/src/qst.cpp#L15012)

this will need to be bumped https://github.com/ZQuestClassic/ZQuestClassic/blob/48660bfcbfad6ef6890913e9e565945e6deb3f4a/src/base/zdefs.h#L153

then the editor needs to expose a checkbox for this new flag

then of course, the `superman=1` lines needs to be removed, and `kickbucket` needs to check (in addition to superman) this new flag

you might want to wait for Emily to give any additional advice or corrections to mine before starting.

ghost:
Added the flag.

src/zc/guys.cpp
```diff
@@ -1743,7 +1743,7 @@ void enemy::leave_item()
 // auomatically kill off enemy (for rooms with ringleaders)
 void enemy::kickbucket()
 {
-	if(!superman)
+	if(!superman && ~flags2&guy_ignorekill)
```
connorjclark:
```suggestion
	if(!superman && !(flags2&guy_ignorekill))
```

idk if previous code is correct or not (order of operations for bitwise logic is a PITA), but I'm certain this is correct.

src/zc/guys.cpp
```diff
@@ -13546,8 +13546,6 @@ void eLilDig::draw(BITMAP *dest)
 
 eBigDig::eBigDig(zfix X,zfix Y,int32_t Id,int32_t Clk) : enemy(X,Y,Id,Clk)
 {
-	superman=1;
-	
 	SIZEflags = d->SIZEflags;
 	if ( ((SIZEflags&guyflagOVERRIDE_TILE_WIDTH) != 0) && txsz > 0 ) { txsz = d->txsz; if ( txsz > 1 ) extend = 3; } //! Don;t forget to set extend if the tilesize is > 1.
```
connorjclark, line 13550: let's also update the other enemies that set superman in their constructors:

```
eMoldorm::eMoldorm eeMOLD
eLanmola::eLanmola eeLANM
eManhandla::eManhandla eeMANHAN
```

(don't forget to add to the `readguys` compat code too)

src/qst.cpp
```diff
@@ -15016,6 +15016,14 @@ int32_t readguys(PACKFILE *f, zquestheader *Header)
...
 				}
 			}
+
+			if (guyversion < 47)
+			{
+				if (tempguy.family == eeDIG && tempguy.misc10 == 1)
+				{
+					tempguy.flags2 |= guy_ignorekill;
+				}
+			}
```
connorjclark, line 15032: there is a big table called `default_guys` in `defdata.cpp` that `readguys` uses to initialize things to the defaults, and then starts overwriting the guys for modern quests. but for `guyversion <= 2`, it just uses whats defined in that big table. It's also used for the "Defaults > Guys" action in the editor to reset guys.

unfortunately this above compat migration code doesn't run for ancient quests (instead `readguys` calls `readherosprites2` and early returns), so we need to handle compat for old stuff in another place. You'll do this inside `init_guys`, probably just need to copy this exact `if` statement you have above. That should give us the default values we want in the editor today, while keeping the old behavior intact for ancient quests.

connorjclark (changes requested):

src/qst.cpp
```diff
@@ -15016,6 +15021,14 @@ int32_t readguys(PACKFILE *f, zquestheader *Header)
 					tempguy.moveflags |= FLAG_CAN_WATERWALK;
 				}
 			}
+
+			if (guyversion < 47)
+			{
+				if (tempguy.family == eeDIG && tempguy.misc10 == 1)
```
connorjclark: shouldn't this be `misc10 != 1` (search for `new eBigDig`)?

Jambu1989: In my test that is not the case

src/zc/guys.cpp
ghost: Do take note those are not the segments but their controllers and have it to prevent memory leaks. Havok would ensue if the controller was destroyed and the segments where still alive. The actual segments themselves don't have it from what I see.

src/qst.cpp
ghost: Yep, taken care of.

ghost:
@connorjclark should be ready to go now

src/zc/guys.cpp
connorjclark: oh, makes sense. i confirmed that Kill All cheat still works with all these enemies so all good.

connorjclark (approved):

connorjclark:
well done 🎉

PR #937, opened 2024-06-27
feat: add shield effect for jinxes, update Whisp Ring

ghost:
Shield Bubbles Added
Whisp Ring updated to have power (jinxdivisor) actually do what is expected with temp jinxes. They were doing absolutely nothing which is honestly unexpected.


src/base/zdefs.h
```diff
@@ -1025,7 +1025,7 @@ enum { e7tNORMAL, e7tTEMPJINX, e7tPERMJINX, e7tUNJINX, e7tTAKEMAGIC, e7tTAKERUPE
      };
 
 // Enemy misc8 types
-enum { e8tSWORD, e8tITEM, e8tALL, e8tLAST };
+enum { e8tSWORD, e8tITEM, e8tBOTH, e8tSHIELD, e8tSHIELDSWORD, e8tSHIELDITEM, e8tALL, e8tLAST};
```
connorjclark: if we end up adding another thing here we'll be entering crazy-ville, so let's turn this into a bitmask enum now

```
enum { e8tSWORD = 1, e8tITEM = 2, e8tSHIELD = 4}; // ALL and LAST aren't used
```

then bump the guy version, and work out a way in this method to coerce older versions to the new flag value, something like:

```
int dm8_jinx = dm8;
if (FFCore.quest_format[vGuys] < ...)
{
	if (dm8 == 0) dm8_jinx = e8tSWORD;
	else if (dm8 == 1) dm8_jinx = e8tITEM;
	else if (dm8 == 2) dm8_jinx = e8tSWORD|e8tITEM;
	else dm8_jinx = 0;
}
```

connorjclark (changes requested):

src/base/zdefs.h
```diff
@@ -1025,7 +1025,7 @@ enum { e7tNORMAL, e7tTEMPJINX, e7tPERMJINX, e7tUNJINX, e7tTAKEMAGIC, e7tTAKERUPE
...
 
 // Enemy misc9 types
```
connorjclark, line 1030: this misc8 value is labeled as "Effect Strength" in the enemy editor... its used for jinx (as you are changing here) and 3 other things in the switch you modified

I don't think we have a way to add a tooltip in this dialog, but we gotta write down what this field does somewhere for users...any ideas? for the least amount of effort, we could make a new file `resources/docs/enemies.txt` and just toss it in there - or there's also `| 4.6b ENEMY EDITOR -> DATA 2  |` in zquest.txt which could be better ...

we gotta  keep this in mind when we re-write the enemy editor

src/qst.cpp
```diff
@@ -15032,6 +15032,27 @@ int32_t readguys(PACKFILE *f, zquestheader *Header)
...
+				{
+					switch (tempguy.misc8) {
+						case 0: //Sword
+							tempguy.misc8 = 1;
+							break;
+						case 1:	//Item
+							tempguy.misc8 = 2;
+							break;
+						case 2: //Both
+							tempguy.misc8 = 3;
```
connorjclark: please use the new enums instead of hardcoding 1/2/3 when setting misc8

src/qst.cpp
```diff
@@ -15032,6 +15032,27 @@ int32_t readguys(PACKFILE *f, zquestheader *Header)
 					tempguy.flags2 |= guy_ignore_kill_all;
 				}
 			}
+
+			if (guyversion < 48)
+			{
+				if (tempguy.family == eeWALK && (tempguy.misc7==e7tPERMJINX || tempguy.misc7==e7tPERMJINX || tempguy.misc7==e7tUNJINX)) //BUBBLE CHECK
```
connorjclark: forgot `e7tTEMPJINX` here (duplicated the perm one). might be the reason for the test failures.

connorjclark (approved):

PR #939, opened 2024-06-27
refactor(zc): clean up jinx handling

ghost:
hero.cpp
HeroClass.ShieldClk
HeroClass.setShieldClk
these are in Zelda.cpp
HeroShieldClk
setShieldClk

Mostly for uniformity across all jinx types. As well a few things in ffcscript.ccp and zc_sys.ccp for some reason


src/zc/ffscript.cpp
```diff
@@ -4314,7 +4314,7 @@ int32_t get_register(int32_t arg)
 		}
 		
 		case HEROSHIELDJINX:
-			ret = Hero.shieldjinxclk * 10000;
+			ret = (int32_t)(Hero.getShieldClk()) * 10000;
```
connorjclark: this cast is redundant, please remove

src/zc/zc_sys.cpp
```diff
@@ -2328,17 +2331,24 @@ int current_item_id(int itype, bool checkmagic, bool jinx_check, bool check_bunn
...
+
+			if (jinx_check && !(itemsbuf[ovid].flags & item_jinx_immune))
+			{
+				// shields
+				if (itemsbuf[ovid].family == itype_shield && HeroShieldClk())
+					return -1;
+				// everything else (including active shields)
+				if (usesSwordJinx(ovid) ? HeroSwordClk() : HeroItemClk())
+					return -1;
+			}
```
connorjclark, line 2338: There is `checkitem_jinx` that distinguishes between sword/item types and uses the corresponding sword/item clk. that method needs to be updated to handle shields

and then i think it can be used here to simplify code, and above at (L2257-2260)

src/zc/zc_sys.cpp
```diff
@@ -2328,17 +2331,24 @@ int current_item_id(int itype, bool checkmagic, bool jinx_check, bool check_bunn
...
 			return ovid;
 		}
 	}
 	auto ret = _c_item_id_internal(itype,checkmagic,jinx_check,check_bunny);
 	if(!jinx_check) //If not already a jinx-immune-only check...
 	{
 		//And the player IS jinxed...
-		if(HeroSwordClk() || HeroItemClk())
+		if(HeroSwordClk() || HeroItemClk() || HeroShieldClk())
```
connorjclark: can you add a `HeroIsJinxed()` please?

src/zc/hero.cpp
connorjclark, line 1: does this block need an update?

https://github.com/ZQuestClassic/ZQuestClassic/blob/97688da502b7f469c8c030916398dd053abfd3c5/src/zc/hero.cpp#L30198-L30203

also search for `HeroItemClk()` - there's a few other places of interest that may need updating

connorjclark (changes requested):

src/zc/hero.cpp
ghost: why yes it does

src/zc/ffscript.cpp
ghost: Done

src/zc/zc_sys.cpp
Jambu1989: `bool HeroIsJinxed()
{
	if (HeroSwordClk() || HeroShieldClk() || HeroShieldClk())
		return true;
	return false;
}`

Abstracted into a function.

ghost: Done

src/zc/zc_sys.cpp
ghost: Done

src/zc/zc_sys.cpp
```diff
@@ -2328,17 +2331,19 @@ int current_item_id(int itype, bool checkmagic, bool jinx_check, bool check_bunn
 			
 			if(checkmagic && !checkmagiccost(ovid))
 				return -1;
-			if(jinx_check && !(itemsbuf[ovid].flags & item_jinx_immune)
-				&& (usesSwordJinx(ovid) ? HeroSwordClk() : HeroItemClk()))
-				return -1;
+
+			if (jinx_check && !(itemsbuf[ovid].flags & item_jinx_immune))
+			{
+				checkitem_jinx(ovid);
```
connorjclark: this isn't returning like it was before

can this be this?

```cpp
if (jinx_check && !checkitem_jinx(ovid))
    return -1
```

src/zc/zelda.cpp
```diff
@@ -1176,6 +1180,14 @@ void setItemClk(int32_t newclk)
 {
     Hero.setItemClk(newclk);
 }
+void setShieldClk(int32_t newclk)
+{
+	Hero.setShieldClk(newclk);
+}
+bool HeroIsJinxed()
+{
+	return (HeroSwordClk() == 0 && HeroItemClk() == 0 && HeroShieldClk() == 0);
```
connorjclark: this is backwards. should be `HeroSwordClk() || HeroItemClk() || HeroShieldClk()`

connorjclark:
can you reword the commit message to be more user-friendly? it should be a user-facing notice on what feature this implemented, and as such it shouldn't mention internal stuff like `shieldjinxclk` (that's more appropriate for refactor commits)

src/zc/zelda.cpp
connorjclark: also the fact this didnt fail any replay gives me some pause. idk what this being wrong does, but have you run these changes and verified they work as expected (i havent)

ghost:
doesn't add anything... shieldjinxclk wasnt fully implemented... and triggers depend on those missing functions as does getcurrentitem when a shield is jinxed

src/zc/zelda.cpp
ghost: Yeah that's a pretty bad bug actually glad you caught it. What this would do is make it so the Hero Is Jinxed 99% of the time... and when that happens it would return the fHighest Inventory Item Ithat is jinx immune making most quest unplayable.

connorjclark: maybe the reason replays didnt fail is selection bias. most people who sent in replays are zelda pros, and so the few quests that have bubbles they probably are too good to be hit by

connorjclark (approved):

PR #940, opened 2024-06-27
feat: add Cure Shield Jinx to bottles

ghost:
Added the Shield Jinx Flag to Bottles.


src/base/misctypes.h
```diff
@@ -80,6 +80,7 @@ struct bottletype
 #define BTFLAG_PERC0       0x01
 #define BTFLAG_PERC1       0x02
 #define BTFLAG_PERC2       0x04
+#define BTFLAG_CURESHJINX  0x08
```
connorjclark, line 83: nice catch (that a bit was skipped)

connorjclark (approved):

PR #941, opened 2024-07-01
feat: add Shield Jinx trigger

ghost:
Shield Jinx Completeness is finished...


connorjclark:
Most changes to the parser will require updating some test snapshots. A catchall script for that is `python tests/update_snapshots.py`. In this case, it's just the test_zscript.py snapshots, so you can just run `python tests/test_zscript.py --update`.

If you haven't set up git lfs yet, see [this](https://github.com/ZQuestClassic/ZQuestClassic/blob/2e0ed2b0407e9686d77823f9a52fcb29c07b49b3/tests/run_replay_tests.py#L6-L9). Not needed if you just run test_zscript.py, but the others need it.

connorjclark (approved):

PR #942, opened 2024-07-02
fix(zc): kill all sound effects for passageways, item cellars, and when stepping out of a warp

ghost:
lets shut these guys up

https://discord.com/channels/876899628556091432/1249794623853957141

### DIFF
--- a/src/zc/hero.cpp
+++ b/src/zc/hero.cpp
@@ -25642,8 +25642,7 @@ bool HeroClass::dowarp(int32_t type, int32_t index, int32_t warpsfx)
 		else                                                  // item room
 		{
 			specialcave = ITEMCELLAR;
-			map_bkgsfx(false);
-			kill_enemy_sfx();
+			kill_sfx();
 			draw_screen(tmpscr,false);
 			
 			//unless the room is already dark, fade to black
@@ -25709,8 +25708,7 @@ bool HeroClass::dowarp(int32_t type, int32_t index, int32_t warpsfx)
 	
 	case wtPASS:                                            // passageway
 	{
-		map_bkgsfx(false);
-		kill_enemy_sfx();
+		kill_sfx();
 		ALLOFF();
 		//play sound
 		if(warpsfx > 0) sfx(warpsfx,pan(x.getInt()));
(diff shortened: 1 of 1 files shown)


src/zc/hero.cpp
```diff
@@ -25709,8 +25708,7 @@ bool HeroClass::dowarp(int32_t type, int32_t index, int32_t warpsfx)
 	
 	case wtPASS:                                            // passageway
 	{
-		map_bkgsfx(false);
-		kill_enemy_sfx();
```
connorjclark, line 25713: why was calling a function named `kill_enemy_sfx` not killing enemy sfx? this is about the patra example you gave right?

ghost: i need to patch this turns out it doesnt work on enemy sfx on enemy in cellars either.
only time it works is when scrolling

ghost: done now also applies when stepping out of item cellars

connorjclark: my question still stands. can you explain why this was broken?

[`kill_enemy_sfx`](https://github.com/ZQuestClassic/ZQuestClassic/blob/985904538f8feccbbf9ec8f92b7669061638d6e0/src/zc/hero.cpp#L25456) loops through all the enemies and stops their sfx. so why was this not sufficient to stop all enemy sfx?

ghost: it's not it shouldn't have been used in the first place... none of the other warps use it. warps should terminate all sounds period.
its only used for scrolling from one screen to the next.

connorjclark: > warps should terminate all sounds period.

I agree with that, so the code looks good to me.

Just trying to understand how this resolves the bug as I understand it. Maybe just the initial PR commit/title is not accurate so thats confused me

this was about the roaring sfx in the room before Ganon still playing even if the player goes back through the passageway, right? I think that isn't an enemy sfx but something else (`bosssfx` on mapscr i think), which would explain the issue if we were not stopping all sfx like a normal screen transition does.

connorjclark (approved):

PR #943, opened 2024-07-04
refactor: Refactoring the ZCMUSIC, ZCMIXER, midi, and midi play

wilsonpe66:
- Updated zcmusic to use string_view and string
 - Updated tests/generate_coverage_report.sh to be executable.
 - Updated script permissions.
 - Changed sound dll to favor a staticly linked library.
 - Fixed ambiguous function class in emscripten.
 - Added an ubuntu dependancy install script.
 - Added namespace wrappers to ZCMUSIC, ZCMIXER, midi, and midi play


connorjclark:
I'm sorry but this is too large of a PR to review with various different concerns. Can you break it up?

.gitignore
```diff
@@ -140,3 +140,8 @@ output/
 changelogs/nightly.txt
 src/base/version_header.h
 venv
+*.log
+scratch
+.idea
+cmake-build-release/
+cmake-build-debug/
```
connorjclark, line 147: `*.log` makes sense (we have allegro.log that our app makes), but for these others you should add them to your  global .gitignore

src/base/zdefs.cpp
```diff
@@ -4,7 +4,7 @@
 #include "base/zapp.h"
 #include "dialog/info.h"
 #include <sstream>
-#include "sound/zcmusic.h"
+#include <sound/zcmusic.h>
```
connorjclark, line 7: our convention is to reserve `<>` imports for std or third party imports

src/dialog/edit_dmap.cpp
```diff
@@ -97,15 +95,9 @@ bool EditDMapDialog::disableMusicTracks()
 
 void EditDMapDialog::silenceMusicPreview()
 {
-	zc_stop_midi();
+	midi::play_midi::stop();
```
connorjclark, line 98: this usage of namespaces is not consistent with anything else in our codebase, so I don't think it's a good idea to introduce.

src/music_playback.h
```diff
@@ -1,9 +1,14 @@
-#ifndef _MUSIC_PLAYBACK_H_
-#define _MUSIC_PLAYBACK_H_
+#pragma once
```
connorjclark, line 1: please defer to existing conventions, we use header guards in pretty much all of these source files

connorjclark (changes requested):
I'm sure there's some things in here that are beneficial, but this PR is too large to be reviewable, and makes some changes that I disagree with. It's ideal to reach out first before sending big PRs like this, especially refactors.

Feel free to hop into our [Discord](https://discord.gg/KrGbjtYD) to discuss things more granularly.

To help find a place to start: what's your main motivation here / what's the thing would you most like to contribute?

src/music_playback.h
wilsonpe66: Putting the guard back are fine. I will also note in the C++ standard that SYMBOLS that start with _ are reserved for the C++ standards library and compiler implmenters so try not to use them as it may break the game :).

wilsonpe66: Thanks for reviewing the pr as far as you did.

src/dialog/edit_dmap.cpp
wilsonpe66: I think, this is good place for a discussion in because in the code base there is little use of encapsulation in the code base.

I am trying to use namesaces as a means of isolating symbols that are being used by other compilation units. Following that I use annymous names in the cpp files to make the symbols local the compilation unit.

If the sybmol is truly global I try to use  inline or constexpr (inplicitly inline) to take advances of the how inline realy means weak linkage in the c++ LOL not realy inline. That change insures that there are no mutiply defined symbols errors in the linker.

I think a discussion is useful to setup a naming convention for the codebase.

Fyi: I am a 10+ super fan of this project and would like share my 20 years of solfware experience to improve the shape of the layout.

Best
Pete

src/base/zdefs.cpp
wilsonpe66: Sure.

.gitignore
wilsonpe66: Sure.

wilsonpe66:
> I'm sorry but this is too large of a PR to review with various different concerns. Can you break it up?

Yeah that is doable.

.gitignore
wilsonpe66: I have another pr that addresses the gitinore changes https://github.com/ZQuestClassic/ZQuestClassic/pull/944

wilsonpe66:
> > I'm sorry but this is too large of a PR to review with various different concerns. Can you break it up?
> 
> Yeah that is doable.

As part of above conversation i have created https://github.com/ZQuestClassic/ZQuestClassic/pull/945 to deal with making zcsound staticly linked.

wilsonpe66:
I will create other smaler prs for the other changes that I have embedded in this pr and not link them here unless you want me link like them here. I will create some discussion about the issues that you have brought up in this pr.

src/music_playback.h
connorjclark: Oh I hadn't noticed that before. They indeed should be changed to the form `FILE_PATH_H_` ([ex](https://source.chromium.org/chromium/chromium/src/+/main:base/allocator/allocator_check.h)). I'll try writing a regex to mass fix these.

wilsonpe66:
Closing this pr as it is too big

PR #945, opened 2024-07-04
refactor: Made the applications staticly linked to zcsound.

wilsonpe66:


connorjclark:
What's the benefit of this being statically linked? The sound library is shared across the editor and the player. Granted the size isn't much (all its dependent sound libs - gme, dumb, etc - are also shared libs), but I don't see why it shouldn't be a shared library.

Aside from that, this PR isn't in a form I'd accept, as it additionally moves stuff around. There is a very high preference for commits doing one thing.

CMakeLists.txt
```diff
@@ -1207,4 +1154,4 @@ if(EMSCRIPTEN)
 	add_subdirectory(web)
 endif()
 
-zc_install_targets(allegro allegro-legacy allegro_with_legacy zlauncher zplayer zeditor zscript zcsound ${SOUNDLIBS})
+zc_install_targets(allegro allegro-legacy allegro_with_legacy zlauncher zplayer zeditor zscript zcsound)
```
connorjclark, line 1157: This would prevent the sounds libs - dumb, gme - from being installed on Linux when using the install command.

wilsonpe66: The rational for this comes when you want to export C++ features.

If you use dynamic linking, you're forced into DLL specifications which basically make it hard to export. Many features that C++ offers if you simply statically link your library.

wilsonpe66: I use Ubuntu Linux as my daily driver and have compiled this code on that machine and it also got past your build process.

wilsonpe66: Read into your initial comment on this I have not tested actually installing the build targets. I will try this and get back to you. Thank you for letting me know.

connorjclark: Oh, so this is motivated by wanting to use `std::string` in the interface?

I'm aware that using standard library objects in a shared library is problematic for shared libraries consumed by third parties - different compiler versions or settings makes things not interoperable at the ABI level - but we are building these libraries for just our own application, so I don't believe this impacts our decision to use a dynamic library here.

The other aspect worth mentioning here is memory allocation - shared libraries must dealloc memory they alloc themselves. This isn't an issue for us either because it's natural to have an explicit freeing method to handle that for sound (`zcmusic_unload_file` in this case).

wilsonpe66: Here are some of the reasons:

- You can export constexpr functions. (see https://en.cppreference.com/w/cpp/language/constexpr).
- You can also export inline definitions in the header files. (So you can better redue the risk of  getting and ready defined symbol ( due to the fact that inline means weak linkage)
- Many data structures member are conextpr supported.
- There a nicer ways to ensure object life times.
- There is support std::string_view which are views in the string like things in including buffers, that point into the - Data -they are passed to as contruction.
- The is std is a great reason use the stl.

constexpr is implicitly inline and noexcept qaulified.

There are more more benifits ....

connorjclark: I think you missed my reasoning why most of the things you list are not precluded by us building as a shared library:

> but we are building these libraries for just our own application, so I don't believe this impacts our decision to use a dynamic library here.

The dll and the consuming application are using the same runtime, as they were compiled together. So it should just work.

But even if not true, I don't believe these benefits are worth the extra binary size.

wilsonpe66:
Oh that is fine. I will just close this or. Thanks for conversation!

Thanks,


*Peter Wilson*

*M:* 646-255-7920

*E: ***@***.***

wilsonpe66:
Closing this pr as it does not match the code culture.

PR #949, opened 2024-07-24
fix(zc): fixed side shielded enemies flags being swapped when moving left/right

ghost:
Fixed the side shields on enemies.


connorjclark:
what was broken?

ghost:
When moving left and right the sides where swapped
So the left side was being treated as the right side side and vice versa.
________________________________

PR #950, opened 2024-07-24
fix(zc): add shadows back to rocks and boulders

ghost:
Seems we are forced you use their own override method to access sprite::drawshadow directly weird... they wont have shadows when they aren't suppose to either. (sideview for example)


connorjclark:
comment on commit title: were shadows drawn for rocks previously?

ghost:
yes

PR #951, opened 2024-07-26
fix(zc): shadows for rocks and boulders not drawing

ghost:


ghost:
Oh the problem here is when things dont move through the Z axis they are not drawing shadows. So come to think of it a lot of other enemies are problably broken as well. Since that's a quest rule. and shadows should be draw regardless of the Z axis if that quest rule is off.

PR #952, opened 2024-07-27
fix(zc): side shield flags on npcs being flipped when facing left/right

ghost:


connorjclark (approved):
lucky us, no replays hit this bug

PR #953, opened 2024-07-27
feat: configurable charge step for enemy rope walk style

ghost:
ropes will use dmisc10 if it is NOT zero or less.

### DIFF
--- a/src/zc/guys.cpp
+++ b/src/zc/guys.cpp
@@ -11602,7 +11602,7 @@ void eStalfos::charge_attack()
 				if(dmisc10<=0 || replay_version_check(0,33))
 					step=zslongToFix(dstep*100)+1;
 				else
-					step=zslongToFix(dstep*100)+(dmisc10/100);
+					step=zslongToFix(dmisc10*100);
 			}
 			else newdir(4,0,0);
 		}
--- a/src/zc/guys.cpp
+++ b/src/zc/guys.cpp
@@ -11599,7 +11599,10 @@ void eStalfos::charge_attack()
 			{
 				dir=ldir;
 				dashing=true;
-				step=zslongToFix(dstep*100)+1;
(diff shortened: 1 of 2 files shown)


connorjclark:
@EmilyV99 Ist das gut?

connorjclark (approved):

src/zc/guys.cpp
```diff
@@ -11599,7 +11599,10 @@ void eStalfos::charge_attack()
 			{
 				dir=ldir;
 				dashing=true;
-				step=zslongToFix(dstep*100)+1;
+				if(dmisc10<=0 || replay_version_check(0,33))
+					step=zslongToFix(dstep*100)+1;
+				else
+					step=zslongToFix(dstep*100)+(dmisc10/100);
```
EmilyV99: Firstly, there's a bug here- `dmisc10/100` is an integer division. So here, any value between "0" and "99" would all do the same thing! Normally in C++, you would avoid this by using floating-point (ex. `dmisc10 / 100.0`) - but for ZC we prefer to keep everything in the fixed-point (`zfix`) typing instead. That is, in fact, already what the rest of this line is doing- so it's simple to convert it to match.
```suggestion
					step=zslongToFix(dstep*100)+zslongToFix(dmisc10*100);
```
or, you can even combine the terms
```suggestion
					step=zslongToFix((dstep+dmisc10)*100);
```

Secondly, personally, I feel like the charge step being customized should *override* the base speed, rather than *add* to it - what if I want to create an enemy that darts around quickly, but then lumbers slowly towards you when it notices you? (...maybe not that necessary a use case, but still it would be simple enough to support the option)
```suggestion
					step=zslongToFix(dmisc10*100);
```

EmilyV99 (changes requested):

EmilyV99 (approved):
Looks good to me with that change then, yeah <3

PR #956, opened 2024-08-02
fix(zq)!: enemy editor throwing errors on exit

ghost:


src/zq/zq_custom.cpp
```diff
@@ -5003,11 +5001,11 @@ void edit_enemydata(int32_t index)
 				int32_t size = 0;
 				((ListData*)enedata_dlg[64+i].dp)->listFunc(-1,&size);
 				// Bound ms[i] as well as enedata_dlg[64+i].d1
```
connorjclark: this comment is now stale, but the code is also self-explanatory, can just delete this comment imo

connorjclark (approved):
I changed the commit title to be `fix(zq)!: ...`, since the `!` omits this from the changelog. This PR fixes a bug introduced in a commit that hasn't seen a release yet, so it's preferred to omit from the changelog.

ghost:
> I changed the commit title to be `fix(zq)!: ...`, since the `!` omits this from the changelog. This PR fixes a bug introduced in a commit that hasn't seen a release yet, so it's preferred to omit from the changelog.

anything else?

PR #961, opened 2024-08-05
feat(zasm): added npc->flags[] npcdata->flags[] as they are now 64bits

ghost:


ghost:
merged with guyflags commit as it should of been processed to begin with.

PR #968, opened 2024-08-08
refactor!: combine guy_flags2 into guy_flags

ghost:
Squashed


Jambu1989:
not possible do to how processors handle bits.

ghost:
@connorjclark  i tried updating the replays didnt work, and the compare isnt working either.

connorjclark:
our upload tool, surge, broke last night. https://github.com/sintaxi/surge/issues/537. I'll pin to a known working version rather than the latest

Updating replays for a refactor is very unexpected, nothing should be changing visually. That should probably be reverted and the reason for the change/regression identified.

ghost:
traps are being weird in those 3 replays
yet they work fine everywhere else,
interestingly enough two of them are still failing which begs the question...
what is really going on here and why...

this refactor doesnt actually change behavior at all so im with connor here

src/zc/guys.cpp
```diff
@@ -1612,7 +1612,7 @@ bool enemy::isOnSideviewPlatform()
 	int32_t usehei = (SIZEflags&guyflagOVERRIDE_HIT_HEIGHT) ? hit_height : 16;
 	if(!get_qr(qr_BROKEN_SIDEVIEW_SPRITE_JUMP)&&fall<0)
 		return false;
-	if(y + usehei >= 176 && currscr>=0x70 && !(tmpscr->flags2&wfDOWN)) return true; //Bottom of the map
+	if(y + usehei >= 176 && currscr>=0x70 && !(tmpscr->flags&wfDOWN)) return true; //Bottom of the map
```
connorjclark: revert

ghost: done

ghost:
finally should be ready to merge now connor

PR #971, opened 2024-08-15
refactor(zq): redesign enemy editor using new GUI

ghost:
Why do my branches keep closing every time i rebase anybody know?


connorjclark:
image

It says you deleted it.

Question: when you rebase, are you pushing and seeing "can't do that, upstream branch have different changes" and then going to GitHub UI and deleting the branch so you can push? If so, you need be doing `git push --force` after a rebase instead.

I could reopen the other but not with this other PR open: 
image
(nothing to do, there's no conversation on the other PR anyhow)

connorjclark:
Also, you could just do a normal merge from main to update, we're gonna squash anyway. Do whatever you're more comfortable with.

src/zc/guys.cpp
```diff
@@ -3799,24 +3799,14 @@ int32_t enemy::takehit(weapon *w, weapon* realweap)
 		//preventing stunlock might be best, here. -Z
 		if(def >= 0) return def;
```
connorjclark: what are these changes? seems unrelated to editor redesign

src/zc_list_data.cpp
```diff
@@ -255,9 +304,14 @@ GUI::ListData GUI::ZCListData::dropsets(bool numbered, bool none)
 	
 	for(int32_t q=0; q < MAXITEMDROPSETS; ++q)
 	{
-		char const* dropname = item_drop_sets[q].name;
+		char* dropname = item_drop_sets[q].name;
 		std::string name;
-		if(numbered)
+		if (dropname[0] == 0)
+		{
+			sprintf(dropname, "zz%03d", q);
```
connorjclark: this is modifying `item_drop_sets[q].name` ... is that intentional?

It'd be better to stick to `fmt::format` (see usage of it just below) - sprintf has issues.

src/zc/guys.cpp
ghost: ... i reverted that,

ghost:
So how we looking now.
I went through everything, and it looks good to my eyes.

connorjclark:
excellent job, I bet this took a lot of effort.

Here's some feedback after using it for a bit.

______

1. The defenses tab had a "Set all" helper that is no longer present 
image

(not suggesting this must be added now, just noting)

2. We have a Flags tab but still put flag stuff elsewhere. For example, "Basics > Size Flags" could be "Flags > Size Flags". Also "Movement".

3. I think removing the nested tabs of Basics would be cleaner, and give us back some vertical space for the most dense tab. Can make a new tab (just after Basics) for "Attributes"

4. Inside "Flags", lets shorten the names of the sub-tabs by not repeating "Flags" (ex: Basic, Behavior, Spawn)

5. I'm not sure about the "Attack" tab given its so minimal, is there room in Effects to put those two fields? Or: if item 3 above gives us back enough vertical space, put it back under Basics? up to you.

6. Halt Rate, Random Rate, Homing Factor, etc... these fields would benefit from a `?` hint

7. all the misc attributes `?` are disabled, is the plan to fill these in later?

8. `Show Script Attrib Metadata` seems unused, and I don't see the use in being able to disable annotating with script labels if present, please remove

9. The `?` hint for "Spawn Animation" says "Animation used while spawning". This is not useful, just remove the `?`. The same is true for Hit Sound, Death Sound.

10. "Shadow Sprite" shows a newline as `\n` for some reason.

11. "Clear" didn't seem to clear everything (Movement flags were still set). "Default" didn't seem to touch Movement flags either (in a New quest file, pressing Default clears all the movement flags).

12. "ASpeed" is repeated twice in "Basics"

ghost:
1. noted
2. done
3. done
4. done
5. explained in discord
6. so can everything else... :D
7. Correct, will be done later. as it's a low priority.
8. Annotation for the Enemy Editor isn't enabled yet, it will be the last thing i add. 
still want me to remove it? Legit copied that tab from the combo editor so it would match.
9. done
10.  done
11. you can check discord now 
12. they are different fields one for Old Animation one for New.

src/dialog/enemyeditor.cpp
```diff
@@ -0,0 +1,1760 @@
...
+	{"Instant",2},
+};
+
+static const GUI::ListData list_deathtype
+{
+	{"Normal",0},
+	{"Explode",1},
+	{"Eight Shards",1}, //Ganon
+	{"Robot Master",2}, //Megaman
+	{"Disinigrate",3}, //BS GANON
```
connorjclark:
```suggestion
	{"Disintegrate",3}, //BS GANON
```

src/dialog/enemyeditor.cpp
```diff
@@ -0,0 +1,1760 @@
...
+		bool doclear = false;
+		AlertDialog("Are you sure?",
+			"Clearing the enemy will set all values to null!",
+			[&](bool ret, bool)
+			{
+				doclear = ret;
+			}).show();
+		if (doclear)
+		{
+			local_guyref = guysbuf[0];
```
connorjclark, line 1666: This can vary based on the qst, right? Would `default_guys[0]` be better?

I did first think `{}` would be best (default ctor) .. but that sets tile width/height to zero instead of 1, which is ick (there's also other fields that "0" isn't the right default for). Setting those fields "real" default values in the guydata struct could make sense but may be more trouble than its worth.

src/dialog/enemyeditor.cpp
```diff
@@ -0,0 +1,1760 @@
...
+		AlertDialog("Are you sure?",
+			"This will reset all values back to their defaults!",
+			[&](bool ret, bool)
+			{
+				doclear = ret;
+			}).show();
+		if (doclear)
+		{
+			local_guyref = default_guys[0];
+			if (index < 177)
```
connorjclark:
```suggestion
			if (index < OLDMAXGUYS)
```

connorjclark:
> Show Script Attrib Metadata seems unused, and I don't see the use in being able to disable annotating with script labels if present, please remove

Ah that makes sense. I've asked Emily to help me understand how that's useful. In the meantime, since we don't have it implemented here, can you just hide the checkbox for now? Just in case other things come up and we end up not implementing it for awhile.

> they are different fields one for Old Animation one for New.

Oh, right! and now you've disabled the irrelevant ones - good stuff.

PR #975, opened 2024-08-24
Fixed launcher not exiting on linux.

wilsonpe66:


connorjclark:
`allegro_init` already installs this as an atexit function.

```
@@void @allegro_exit();
@xref install_allegro, allegro_init, destroy_bitmap
@eref ex3d, exscn3d, exswitch, exxfade, exzbuf
@shortdesc Closes down the Allegro system.
   Closes down the Allegro system. This includes returning the system to
   text mode and removing whatever mouse, keyboard, and timer routines have
   been installed. You don't normally need to bother making an explicit call
   to this function, because allegro_init() installs it as an atexit()
   routine so it will be called automatically when your program exits.

   Note that after you call this function, other functions like
   destroy_bitmap() will most likely crash. This is a problem for C++ global
   destructors, which usually get called after atexit(), so don't put Allegro
   calls in them. You can write the destructor code in another method which
   you can manually call before your program exits, avoiding this problem.
```

I checked on Mac in a debugger and `allegro_exit` is indeed getting called when main returns.

Can you provide more information? Did this change really make a difference for you?

wilsonpe66:
While that may be true. Exit is not being properly call on unbuntu linux 24.04. This change ensures that is done so allegro can exit properly and terminate the application with status code 0.

wilsonpe66:
Without this change the application refuses to dei.

connorjclark:
I believe that indicates that some other atexit register function is behaving abnormally- if one returns non zero then no other exit handlers run https://stackoverflow.com/a/5674644/2788187

This change is fine (but please change it to just call allegro_exit() directly), but it'd be great to learn what handler is failing.

wilsonpe66:
I will change later tonight. I and I can try to find a cause for that issue as well.

Best

connorjclark:
Cool, good luck! No problem if you can't. Thanks for the fix.

wilsonpe66:
In the main branch 

on exit attempt

void allegro_exit(void) is being called however it is freezing on function

There is the call stack for thread 1 ( the main thread)

```
 - glReadPixels(x, gl_y, w, h,
         get_glformat(format, 2),
         get_glformat(format, 1),
         ogl_bitmap->lock_buffer);
  - ok = ogl_lock_region_nonbb_readwrite_fbo(bitmap, ogl_bitmap,
         x, gl_y, w, h, format);
 - ok = ogl_lock_region_nonbb_readwrite(bitmap, ogl_bitmap,
            x, gl_y, w, h, format, &restore_fbo);
 - lr = bitmap->vt->lock_region(bitmap, xc, yc, wc, hc, format, flags);
 - return al_lock_bitmap_region(bitmap, 0, 0, bitmap->w, bitmap->h, format, flags);
 - if (!(src_region = al_lock_bitmap(src, lock_format, ALLEGRO_LOCK_READONLY)))
 - if (!transfer_bitmap_data(bitmap, clone)) {
 - clone = al_clone_bitmap(bitmap);
 - al_convert_bitmap(bitmap);
 - _al_convert_to_memory_bitmap(b);
 - convert_display_bitmaps_to_memory_bitmap(d);
 - display->vt->destroy_display(display);
 - al_destroy_display(_a5_display);
 - _a5_destroy_screen();
 -  gfx_driver->exit(screen);
 - return _set_gfx_mode(card, w, h, v_w, v_h, TRUE);
 - set_gfx_mode(GFX_TEXT, 0, 0, 0, 0);
 - (*func)();
 - allegro_exit(); 
```


My guess is that there must the a lock aquire attempt being hung up on.

wilsonpe66:
I am have push the change and I am just waiting for build testing to complete.

wilsonpe66:
On review, it looks like by added my call to allegro_exit() it breaks the lock on linux.

wilsonpe66:
The editor has the same issue. will make a pr that issue to be resolved as well.

connorjclark:
There's actually multiple places where the program exits and this should probably be called in all of them, including in the player and the editor. I went ahead and refactored places that exit to use a common function, so this issue should be fixed in more places with that approach. I'll go ahead and close these PRs (I'll merge my fix soon).

Thanks for reporting this. I shared the root cause (the opengl hang when the display is destroyed) upstream: https://github.com/liballeg/allegro5/issues/1575

PR #976, opened 2024-08-25
fix: Fixed edit exit not exiting on linux.

wilsonpe66:


connorjclark:
https://github.com/ZQuestClassic/ZQuestClassic/pull/975#issuecomment-2309025896

PR #980, opened 2024-08-28
fix: Fixed png logo not loading problem with linux.

wilsonpe66:


connorjclark:
This was a build configuration bug, fixed in https://github.com/ZQuestClassic/ZQuestClassic/commit/c0553e6a9880b16f54231db6fa931dd5161a2cf5

PR #986, opened 2024-09-03
refactor(zq): upgrade enemy editor's enemy selector to new gui

ghost:
old:

image

new:

image

### DIFF
--- a/src/dialog/info_lister.cpp
+++ b/src/dialog/info_lister.cpp
@@ -426,10 +426,10 @@ void EnemyListerDialog::update()
 		guydata const& enemy = guysbuf[selected_val];
 		widgInfo->setText(fmt::format(
 			"#{}\nTile: {}\nsTile: {}"
-			"\neTile: {}\nHP: {}\nDamage:\nFamily: {}\nDrop: {}\nScript: {}\nW Script: {}"
+			"\neTile: {}\nHP: {}\nDamage: {}\nW. Damage: {}\nFamily: {}\nDrop: {}\nScript: {}\nW Script: {}"
 			"\n\nCopied:\n{}",
 			selected_val, enemy.tile, enemy.s_tile,
-			enemy.e_tile, enemy.hp, enemy.dp, enemy.family, enemy.item_set, enemy.script, enemy.weaponscript,
+			enemy.e_tile, enemy.hp, enemy.dp, enemy.wdp, enemy.family, enemy.item_set, enemy.script, enemy.weaponscript,
 			copied_name));
 		widgPrev->setDisabled(false);
 		if (get_qr(qr_NEWENEMYTILES))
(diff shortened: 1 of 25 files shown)


src/zq/zq_custom.cpp
```diff
@@ -2881,44 +2881,7 @@ void elist_rclick_func(int32_t index, int32_t x, int32_t y)
...
-		}
-		else
-		{
-			if(index != 0)
-			{
-				edit_enemydata(index);
-			}
-		}
-		
-		index = select_enemy("Select Enemy",index,true,true,foo);
```
connorjclark, line 2919: should the other `select_enemy` get replaced with this new one now? then we can delete `select_enemy`

if thats tricky then nvm for now

src/dialog/info_lister.cpp
```diff
@@ -383,3 +384,151 @@ void SubscrWidgListerDialog::update()
...
+		Z_error("Could not write to .znpc packfile %s\n", temppath);
+		InfoDialog("ZNPC Error", "Could not save the specified NPC.").show();
+	}
+	pack_fclose(f);
+}
+bool EnemyListerDialog::load()
+{
+	if (selected_val < 0)
+		return false;
+	if (!prompt_for_existing_file_compat(fmt::format("Load NPC (replacing '{}' #{}) (.z)", guy_string[selected_val], selected_val).c_str(), "znpc", NULL, datapath, false))
```
connorjclark:
```suggestion
	if (!prompt_for_existing_file_compat(fmt::format("Load NPC (replacing '{}' #{}) (.znpc)", guy_string[selected_val], selected_val).c_str(), "znpc", NULL, datapath, false))
```

ghost:
image
image2

src/zc_list_data.cpp
```diff
@@ -198,12 +198,10 @@ GUI::ListData GUI::ZCListData::enemies(bool numbered, bool defaultFilter)
...
 		if(defaultFilter)
 		{
-			if(q >= 11 && q <= 19) //Segment components
+			if(q >= 0 && q <= 19) //Room Guys
 				continue;
 			if(q < OLDMAXGUYS && old_guy_string[q][strlen(old_guy_string[q])-1]==' ')
 				continue; //'Hidden' enemies
-			if(guysbuf[q].family == eeGUY)
-				continue; //Guys
 		}
```
connorjclark, line 205: so the first 20 are room guys, and now this skips them explicitly

and i see other eeGUYs in defdata.cpp but they are still excluded because their names end in a space (like eBATROBEKING whose name is `"Wizzrobe (Bat 2) "`)

what additional enemies get shown now w/ this change?

ghost: that commit fixes the issue with guys 256 to 511 not showing up which are in fact eeGUY by default...

connorjclark: ah nice!

connorjclark (approved):

PR #989, opened 2024-09-08
refactor(zq): upgrade screen enemy selector to new gui

ghost:
image

image


ghost:
should be ready now.

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+			}
+		}
+		if (!confirm) rerun_dlg = true;
+		return true;
+	}
+	case message::CANCEL:
+		memcpy(thescr->enemy, oldenemy, 10 * sizeof(word));
+		return true;
+	}
+	rerun_dlg = refresh;
```
EmilyV99:
```suggestion
	if(refresh) rerun_dlg = true;
```
to match the `if(!confirm) rerun_dlg = true;` above; and also probably cleaner in general to not set rerun_dlg to anything but `true`, as nothing ever should be writing it to false except the dialog runner code (in this case, it should always ALREADY be false when you write it to false, so like, this *really* doesn't matter that much.

src/dialog/enemypattern.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+	window = Window(
+		title = "Select Pattern",
+		info = "These will change how the screen enemies will spawn.",
+		shortcuts = {
+			K = message::OK
+		},
+		onClose = message::CANCEL,
+		Column(hAlign=1.0,
+			Label(text="Classic: Psuedorandom locations near"
```
connorjclark:
```suggestion
			Label(text="Classic: Pseudorandom locations near"
```

src/dialog/enemypattern.cpp
```diff
@@ -0,0 +1,80 @@
...
+				"\nthe middle of the screen."),
+			Label(text="Random: Any available location"
+				"\nat a sufficent distance from the player."),
```
connorjclark:
```suggestion
				"\nat a sufficient distance from the player."),
```

src/dialog/enemypattern.cpp
```diff
@@ -0,0 +1,80 @@
...
+				onSelectFunc = [&](int32_t val)
+				{
+					last_pattern = val;
+				}),
+			Row(
+				topPadding = 0.5_em,
+				vAlign = 1.0,
+				spacing = 2_em,
+				Button(
+					text = "O&K",
```
connorjclark, line 51: why not `&OK`?

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+					{
+						confirm = ret;
+					}).show();
+			}
+		}
+		if (!confirm) rerun_dlg = true;
+		return true;
+	}
+	case message::CANCEL:
+		memcpy(thescr->enemy, oldenemy, 10 * sizeof(word));
```
connorjclark:
```suggestion
		memcpy(thescr->enemy, oldenemy, sizeof(thescr->enemy));
```

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+void call_screenenemies_dialog()
+{
+	ScreenEnemiesDialog().show();
+}
+
+ScreenEnemiesDialog::ScreenEnemiesDialog() :
+	thescr(Map.CurrScr())
+{
+	memcpy(oldenemy, Map.CurrScr()->enemy, 10 * sizeof(word));
```
connorjclark:
```suggestion
	memcpy(oldenemy, Map.CurrScr()->enemy, sizeof(oldenemy));
```

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+{
+	using namespace GUI::Builder;
+	using namespace GUI::Props;
+	using namespace GUI::Key;
+
+	window = Window(
+		use_vsync = true,
+		title = "Select Enemies",
+		info = "Click the ten spaces to the left to select different enemies to spawn on this screen."
+		" You can also access Enemy Flags and set the Spawning Pattern.",
```
connorjclark:
```suggestion
		"\nYou can also access Enemy Flags and set the Spawning Pattern.",
```

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+			),
+			Column(
+				widgPrev = TileFrame(visible = false),
+				widgInfo = Label(text = "")
+			)
+		)
+	);	
+	RebuildList();
+	UpdatePreview();
+	return window;
```
connorjclark, line 187: Do we have a way to persist the selected index when the user comes back to the dialog (like by closing the Select Enemy dialog, or closing Flags/Pattern, etc)?

src/zq/zquest.cpp
```diff
@@ -18372,6 +18331,7 @@ static DIALOG enemy_dlg[] =
 
 int32_t onEnemies()
 {
+	/*
```
connorjclark: remember to delete this, and enemy_dlg, and select_enemy

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+		bool end = false;
+		bool confirm = true;
+		for (int32_t i = 0; i < 10; i++)
+		{
+			if (thescr->enemy[i] == 0)
+				end = true;
+			else if (end)
+			{
+				AlertDialog("Inactive Enemies!",
+					"Enemies won't appear if they're preceded by '(None)' in the list! Continue?",
```
connorjclark:
```suggestion
					"Enemies won't appear if preceded by '(None)' in the list! Continue?",
```

src/dialog/screen_enemies.cpp
```diff
@@ -0,0 +1,260 @@
...
+					[&](bool ret, bool)
+					{
+						confirm = ret;
+					}).show();
```
connorjclark, line 257: this needs a break or something, because the same prompt will keep coming up multiple times if you have a `None` followed by multiple enemies

src/dialog/enemypattern.cpp
connorjclark, line 1: You can't tell what the selected value for the pattern is. Maybe let's add it to the screen_enemies dialog?

Also, it should pre-select the current pattern value when you open this dialog, but right now it selects the first one always.

src/dialog/screen_enemies.cpp
connorjclark, line 1: Copying the pasting sets the currently selected index back to one (it refreshes the entire dialog)

connorjclark, line 1: If I copy an enemy with a really long name, the "Copied: " text is cut off until the dialog is refreshed:

image

image

But it shouldn't get bigger at all. This text should wrap, the dialog shouldn't change size on you.

Or maybe the currently copied enemy can just start with `"[COPIED]"` or `* ` in the list? Seems to be plenty of room for that.

connorjclark (changes requested):

src/dialog/enemypattern.cpp
ghost: cause the Hot Key for closing dialogs is K not O and has been that way for decades

src/dialog/screen_enemies.cpp
ghost: @EmilyV99

src/zq/zquest.cpp
ghost: 🫡

src/dialog/screen_enemies.cpp
ghost: yes it needs to refresh the dialog afterall

src/dialog/screen_enemies.cpp
EmilyV99: If the dialog is "refreshed", it does re-open the dialog, yes; but it does NOT call the constructor again. Any member variables in the ScreenEnemiesDialog will persist through the refresh. `view()` will overwrite any of them related to the actual dialog, as it constructs a new underlying allegro `DIALOG[]`, but the actual higher-level ScreenEnemiesDialog can persist any information it wants through a refresh. (So, just store the current index in a member variable, set to a default in the *constructor*, and used as the starting index of the lister in `view()`)

src/dialog/screen_enemies.cpp
EmilyV99: > yes it needs to refresh the dialog afterall

which should not change the index. See my other comment.

src/dialog/enemypattern.cpp
ghost: 1. You cant tell hat the enemy flags are either. they are completely different dialogs.
2. Easy fix, give me a sec

src/dialog/screen_enemies.cpp
ghost: roger

src/dialog/screen_enemies.cpp
EmilyV99: > Or maybe the currently copied enemy can just start with `"[COPIED]"` or `* ` in the list? Seems to be plenty of room for that.

Something can be copied and no longer in the list, is the issue.

ghost: fixed

src/dialog/enemypattern.cpp
connorjclark, line 1: You can copy `(None)` but you can't paste it.

No preference on how to resolve this. Though it did make me thing that BACKSPACE would be nice for deleting an enemy.

connorjclark, line 1: When you open the screen_enemies dialog, and press down on the keyboard twice, you can no longer double click any entry to open the enemy selector. I managed to get out of this state by pressing ENTER and going back.

connorjclark (changes requested):

src/dialog/enemypattern.cpp
ghost: Delete Key will clear a field. Either way, this needs to be addressed.

ghost: if we really wanted too we can add backspace and have that clear all enemies of the same type

ghost: @EmilyV99 no idea how to resolve this one.

EmilyV99: no idea either

EmilyV99: Why can't you paste none? are you using 0 as the "nothing is copied" value? just use -1 instead or something

connorjclark: Backspace would be nice. I don't have del on some of my keyboards.

ghost: fixed

ghost: this might be an issue with the widgets themselves. same thing happens with all the InfoListers with their DoubleClick breaking.

connorjclark (approved):
Great work!

src/dialog/enemypattern.cpp
connorjclark: Cool, existing bug. Let's land this.

PR #990, opened 2024-09-08
feat: Better Keese Movement

ghost:
New QR Added QR Makes it so their Z Variance is multiplied by step/dstep
We dont have to bump GuyVersion either :p


ghost:
oh btw this fixes this bug here.
https://discord.com/channels/876899628556091432/1246492945490641098/1246492945490641098

Jambu1989:
ill be deleting this account at the end of the week @connorjclark 
you either merge it or loose the changes.

PR #991, opened 2024-09-10
feat: make the special sfx of enemy weapons configurable too

ghost:


src/zc/weapons.cpp
```diff
@@ -1191,6 +1193,21 @@ void weapon::eweapon_overrides()
 		}
 	}
 	else step = e->wstep;
+	if (e->firesfx < 0)
```
connorjclark:
```suggestion
	if (e->firesfx <= 0)
```
?

connorjclark: but also, this block of code just seems to repeat what is already happening in the weapon constructor switch, why is it needed?

src/zc/guys.cpp
```diff
@@ -11539,11 +11549,6 @@ void eStalfos::KillWeapon()
 			}
 		}
 	}
-	
-	if(wpn==ewBrang && !Ewpns.idCount(ewBrang))
-	{
-		stop_sfx(WAV_BRANG);
-	}
```
connorjclark, line 11546: wouldn't this keep the sfx playing for a tiny bit even when all dead?

should this be `stop_sfx(specialsfx)` now (though specialsfx is still always the same for any specific enemy class, but I guess you plan to make that configurable)

ghost: that codeblock does nothing.

src/zc/weapons.cpp
ghost: script created weapons just use the constructor.
enemy created weapons use eweapon_overrides()

src/zc/guys.cpp
connorjclark: Explain please - is there another part that stops the sfx early when a weapon is deleted?

ghost: yes the weapon deconstructor does it

src/zc/weapons.cpp
ghost: this is actually incorrect.. should be setting specialsfx to firesfx its been fixed

src/base/qst.cpp
```diff
@@ -9779,14 +9779,20 @@ static void guy_update_firesfx(guydata& tempguy)
 					tempguy.firesfx = WAV_ZN1ICE;
 					break;
 				case ewRock:
-					if (get_qr(qr_MORESOUNDS)) tempguy.firesfx = WAV_ZN1ROCK;
```
connorjclark, line 9782: so this is only fine b/c `enemy::wpnsfx` is still checking the qr. hm ok.

connorjclark (approved):

PR #993, opened 2024-09-14
Feat: Effects Tab expanded to Enemy Editor

ghost:


ghost:
for some reason moving death effects breaks RNG completely... so yeah screw this. i've tried everything i can to get this to work at no avail.

EmilyV99:
> for some reason moving death effects breaks RNG completely... so yeah screw this. i've tried everything i can to get this to work at no avail.

If you change even a single thing that changes a single call to any rng, it desyncs all the RNG. The calls need to happen in the ***EXACT*** same order, or it breaks, by definition. This *CAN* be fixed with a replay version check, though, if you can let it still run the old code for replays- but you want to narrow down what old code NEEDS to run in that case.

It looks like there's a few separate functions here that you tried to handle moving (DeathFX, TrySplitting, TryTribble, update_eathero, eathero); why don't you try doing those as separate commits and seeing which one(s) cause the trouble?

ghost:
good point i didn't think to call these in the same order.

ghost:
scraped, this has been an nightmare to implement and i've completely lost interest.

PR #995, opened 2024-09-26
refactor(zq): upgrade midi editor to new gui

ghost:
Before

image

image

After

image

image


src/dialog/midieditor.cpp
```diff
@@ -0,0 +1,329 @@
...
+{
+	using namespace GUI::Builder;
+	using namespace GUI::Props;
+	using namespace GUI::Key;
+	zc_stop_midi();
+	get_midi_info(local_midiref.data, &Midi_Info);
+	timestring = timestr(Midi_Info.len_sec);
+	window = Window(
+		use_vsync = true,
+		onTick = [&]() {lposition->setText(midi_pos >= 0 ? to_string(midi_pos) : "-1"); return ONTICK_CONTINUE;},
```
connorjclark, line 41:
```suggestion
		onTick = [&]() {lposition->setText(midi_pos >= 0 ? to_string(midi_pos) : "0"); return ONTICK_CONTINUE;},
```

feels better to me to show 0 if not playing

src/dialog/midieditor.cpp
```diff
@@ -0,0 +1,329 @@
...
+					minwidth = 60_px,
+					type = GUI::Button::type::ICON,
+					icon = BTNICON_ARROW_RIGHT3,
+					onClick = message::FFF)
+			),
+			Rows<4>(vAlign = 1.0,
+				Label(text = "MIDI:", rightPadding = 0_px, textAlign = 1.0),
+				Label(text = (local_midiref.data) ? "Loaded" : "Invalid", rightPadding = 0_px, textAlign = 1.0, disabled = !(local_midiref.data)),
+				Label(text = "Position:", rightPadding = 0_px, textAlign = 1.0),
+				lposition = Label(text = "-1", fitParent = true, rightPadding = 0_px, textAlign = 1.0),
```
connorjclark, line 79:
```suggestion
				lposition = Label(text = "0", fitParent = true, rightPadding = 0_px, textAlign = 1.0),
```

connorjclark:
what do you think should happen when Play is pressed when already playing?

the old dialog stuttered for a split second

currently this new one seems to jump to the last note (but sometimes I think it actually does restart from the beginning)

can we get it to restart the midi when pressing play again?

ghost:
the message handler is the same as the one from before.
if its not working the same that means i made a mistake.

src/dialog/midieditor.cpp
ghost: feels better to me to have it work the same way as it did before tbh. which was -1 i can change it if you really want me to

EmilyV99: Disagree for consistency sake; this has always indicated `-1` as not playing

src/dialog/midieditor.cpp
EmilyV99: see above, Disagree for consistency sake; this has always indicated `-1` as not playing

src/dialog/midieditor.cpp
connorjclark: -1 makes sense from a programmer's perspective. I don't think a user really has the same perspective though.

I don't feel strongly about it. up to you @Jambu1989

PR #996, opened 2024-09-26
refactor(zq): upgrade sfx editor to new gui

ghost:
Before

image

image

After

image

image


src/zq/zquest.cpp
connorjclark, line 1: let's delete sfx_edit_dlg, sfxlist_dlg, and select_sfx

ghost: done

PR #1004, opened 2024-10-18
feat: logo image in 1x zoom scale

denilsonsa:
The original `ZC_Logo.png` was scaled up unevenly. Some columns/rows are repeated 10 pixels, others are repeated 9 pixels. And there are even a few rows/columns that are unique (i.e. one thin strip of the wrong color).

Here, I've restored the logo back to the 1x zoom scale. I've done it by using the following Python script, followed by a little bit of manual cleanup.

```python
from pathlib import Path
from PIL import Image

with Image.open(Path('~/Downloads/ZC_Logo.png').expanduser()) as img:
    next_x = 0
    duplicates = 0
    prev_line = None

    for x in range(img.width):
        line = img.crop((x, 0, x + 1, 0 + img.height))
        if line == prev_line:
            duplicates += 1
        else:
            img.paste(line, (next_x, 0))
            next_x += 1
            if prev_line is not None:
                print("{0} duplicate columns".format(duplicates))
            duplicates = 0
        prev_line = line

    next_y = 0
    duplicates = 0
    prev_line = None
    for y in range(img.height):
        line = img.crop((0, y, 0 + img.width, y + 1))
        if line == prev_line:
            duplicates += 1
        else:
            img.paste(line, (0, next_y))
            next_y += 1
            if prev_line is not None:
                print("{0} duplicate rows".format(duplicates))
            duplicates = 0
        prev_line = line

    img.crop((0, 0, next_x, next_y)).save("out.png")
```

Having the base 1x logo will be easier to modify/tweak in the future, and easier to scale up to whatever dimension needed.

If you want, we can also replace the `ZC_Logo.png` file with a properly scaled up version. The dimensions will be slightly different, though: 2490x900 instead of 2365x865.

We could also reuse the same Python script to generate 1x versions of the `ZC_Icon_*.png` icons.


connorjclark:
hello! Thanks for the attention to detail

fwiw the original photoshop file is [here](https://github.com/ZQuestClassic/commons/blob/main/ZC%20Logo%20%26%20Icon/ZC_Logo.psd) - I think it'd be best to edit that file directly (done by someone with pixel art experience, ie. not me 😄), and re-export the image as PNG at various scales as needed.

i put the old and 1x one side-by side, blown up with `image-rendering: pixelated` (that's super important else the 1x variant loses visual fidelity) and can't really any difference... but I believe you if you say its not pixel perfect.

image

denilsonsa:
How to reproduce:

1. Open `ZC_Logo.png` in Gimp.
2. View → Show Grid.
3. Image → Configure Grid.
4. Now try to find a grid that perfect matches the scaled up pixels. (Spoiler alert: you can't.)

10x10 grid

9x9 grid

If you zoom in enough, you can see some pixels are rectangular instead of square. Which you can measure by either using the Measure Tool, or by making a selection with the Rectangle Select Tool.

---

If you think we should replace the original file with this new x1 file, I can go ahead and do it. I'm sorry, but I haven't tested the implications of doing so. Will it be rendered correctly in-game? Will it be blurry? I don't know.

Alternatively, we can keep the 1x version separate from a new (properly scaled) 10x version. Your choice. (Or a new 2x version, or whatever-x version.)

denilsonsa:
> fwiw the original photoshop file is [here](https://github.com/ZQuestClassic/commons/blob/main/ZC%20Logo%20%26%20Icon/ZC_Logo.psd)

Wait, what? I didn't know we had this repository! https://github.com/ZQuestClassic/commons

(But also I'm quite new to this project, and it all started because I wanted to submit [some images to SteamGridDB](https://www.steamgriddb.com/game/5467934).)

Maybe that's an "oops" from my side. In any case, I'll have to take a look at it later. "Right now" I'm in a hurry, and I just wanted to give you that quick-ish reply.

denilsonsa:
FYI, I've opened a PR to the commons repository: https://github.com/ZQuestClassic/commons/pull/1

PR #1010, opened 2024-10-28
fix(zq): added unused enemy slots

Jambu2024:
Guys and Item Fairy shall remain hidden. As editing those breaks stuff.


src/defdata.cpp
```diff
@@ -521,7 +521,7 @@ guydata default_guys[OLDMAXGUYS] =
...
 	// eBATROBEKING
-	{ (guy_flags)0x0000000000000000,        0,     0,    0,      0,    0,    0,        0,    16,     1,      0,    eeGUY,            0,    aNONE,         aNONE,             0,      0,     0,     0,    wNone,           0,     0,      0,      0,    0,    isNONE,            {0,              0,                0,               0,           0,    0,    0,      0,    0,    0,    0,    0,    0,    0,    0,  },   -1,                -1,            0,    { edSTUNORCHINK,0,0,edIGNORE,edIGNORE,edIGNORE,edCHINK,edSTUNONLY,0,0,0,0,0,0,0,0,edIGNORE,0,0 }, 0, WAV_EDEAD, },
+	{ (guy_flags)0x0000000000000000,      169,     4,    0,      0,    0,    0,     4960,    16,     3,     28,    eeWIZZ,           9,    aWIZZ,         aNEWWIZZ,        256,    256,     4,    16,    ewMagic,         0,     0,      0,      0,    0,    isMAGICBOMBS,      {0,              2,                eBAT,            0,          74,    0,    0,      0,    0,    0,    0,    0,    0,    0,    0,  },   -1,                -1,            0,    { edSTUNORCHINK,0,0,edIGNORE,edIGNORE,edIGNORE,edCHINK,edSTUNONLY,0,0,0,0,0,0,0,0,edIGNORE,0,0 }, 0, WAV_EDEAD, },
 	// eGLEEOK1F
```
connorjclark, line 525: I tried spawning this enemy, but it didn't do anything.

image


Same for the ceiling/floor masters / grapbugs.

src/zq/zq_misc.cpp
```diff
@@ -427,35 +427,35 @@ const char *old_guy_string[OLDMAXGUYS] =
...
     // 140
-    "Octorok (L4, Slow)", "Octorok (L4, Fast)", "Trap (8-Way) ", "Trap (Diagonal) ", "Trap (/, Constant) ",
+    "Octorok (L4, Slow)", "Octorok (L4, Fast)", "Trap (8-Way)", "Trap (Diagonal)", "Trap (/, Constant)",
     // 145
-    "Trap (/, Line of Sight) ", "Trap (\\, Constant) ", "Trap (\\, Line of Sight) ", "Trap (CW, Constant) ", "Trap (CW, Line of Sight) ",
+    "Trap (/, Line of Sight)", "Trap (\\, Constant)", "Trap (\\, Line of Sight)", "Trap (CW, Constant)", "Trap (CW, Line of Sight)",
     // 150
-    "Trap (CCW, Constant) ", "Trap (CCW, Line of Sight) ", "Wizzrobe (Summoner)", "Wizzrobe (Ice) ", "Shooter (Magic)",
+    "Trap (CCW, Constant)", "Trap (CCW, Line of Sight)", "Wizzrobe (Summoner)", "Wizzrobe (Ice)", "Shooter (Magic)",
     // 155
```
connorjclark, line 459: These traps don't currently work as enemy spawns, that is why they are hidden. They only work if placed as a flag (or the trap combo types). There's not even any flags or combo types for the diagonal variants, so I think that isn't implemented either. We shouldn't display these in the enemy list until they work.

Also realizing now that these traps (which are at least implemented, but don't work as enemy spawns) are currently visible today, but probably should be made hidden until we make them actually work- 
image

connorjclark (commented):
The enemies changed from hidden to visible are not implemented, so why would we want to display them in the enemy lister?

Jambu2024:
cause people are bitching about it, if you don't want to add it that's find with me. we can mark the bug report as denied

PR #1012, opened 2024-10-30
fix(zc): triggered autowarp combos use sidewarps, screenflags for direct autowarps and sidewarps work when respective combo is triggered.

Jambu2024:


src/zc/combos.cpp
```diff
@@ -968,12 +968,23 @@ bool trigger_warp(newcombo const& cmb)
 	mapscr* wscr2=&TheMaps[ws];
 	auto wx = wscr2->warpreturnx[wrindex];
 	auto wy = wscr2->warpreturny[wrindex];
```
connorjclark:
```suggestion
	int wx = wscr2->warpreturnx[wrindex];
	int wy = wscr2->warpreturny[wrindex];
```

src/zc/hero.cpp
```diff
@@ -24487,6 +24487,8 @@ void HeroClass::checkspecial2(int32_t *ls)
 		bool hasStep[4] = {false};
 		for(auto p = 0; p < 4; ++p)
 		{
+			if ((z > 0 || fakez > 0) && !(tmpscr->flags2 & fAIRCOMBOS))
+				break;
```
connorjclark: is this related to warps?

src/zc/combos.cpp
```diff
@@ -968,12 +968,23 @@ bool trigger_warp(newcombo const& cmb)
...
+	if (stype == cAWARPA) //these use sidewarps
+	{
+		wtype = wscr->sidewarptype[index];
+		if (wtype == wtNOWARP)
+			return false;
+		wsfx = cmb.attribytes[0];
+		tdm = wscr->sidewarpdmap[index];
+		tscr = wscr->sidewarpscr[index];
+		ws = (DMaps[tdm].map * MAPSCRS + wscr->sidewarpscr[index] + DMaps[tdm].xoff);
+		wscr2 = &TheMaps[ws];
```
connorjclark, line 981: wscr2 isn't used after this line, did you expect it to be somehow?

src/zc/hero.cpp
Jambu2024: yes it causes the combo to effect link when in air

src/zc/combos.cpp
Jambu2024: fixed

src/zc/combos.cpp
Jambu2024: these are not ints

connorjclark: but `wx = Hero.getX().getInt();` is - so i'm worried about under/overflow, which could happen now and may happen in z3.

src/zc/combos.cpp
```diff
@@ -968,12 +968,25 @@ bool trigger_warp(newcombo const& cmb)
...
+		wx = wscr2->warpreturnx[wrindex];
```
EmilyV99, line 982: This whole if needs to be ABOVE the check for direct warp, because you're just setting wx/wy here (overwriting the direct warp above)

src/zc/combos.cpp
EmilyV99: Agree with Connor here.

src/zc/hero.cpp
EmilyV99: This is step combos, not warps? It is a correct bugfix I believe, but not related to warps.

This would appear to be a correct bugfix though...for [this bug](https://discord.com/channels/876899628556091432/1275994756266197053) which was already marked "closed"?

Jambu2024: triggers tab to be more specific.

Jambu2024:
changelog summary of everything in first post.

PR #1016, opened 2024-11-01
fix(zq): fix tile protection for darknut, gleeok, ganon, and patra enemies

Jambu2024:


connorjclark (commented):
1. Can you provide a better title describing the fix - what was wrong? needs a better message for changelog purposes, and for me to know how to test/verify this PR
2. i don't know what is happening outside the zq_tiles.cpp changes, is that related to the tile protection changes?

EmilyV99:
>     2. i don't know what is happening outside the zq_tiles.cpp changes, is that related to the tile protection changes?

yes, that's what manages the tile protection popup. Apparently some of the "hitboxes" were wrong, so it wasn't popping up in the right spots for some enemies?

Jambu2024:
link to bugreport
https://discord.com/channels/876899628556091432/1279189950008524878/1279189950008524878

to verify copy a tile to Darknut Broken Shield Tiles, Gleeok Tiles (theirs a ton here my god), Patra Inner Eyes, and Ganons Tiles.

PR #1017, opened 2024-11-02
fix(zc): remove ladder during warp

Jambu2024:


src/zc/hero.cpp
```diff
@@ -25490,7 +25490,8 @@ bool HeroClass::dowarp(int32_t type, int32_t index, int32_t warpsfx)
 	// Drawing commands probably shouldn't carry over...
 	if ( !get_qr(qr_SCRIPTDRAWSINWARPS) )
 		script_drawing_commands.Clear();
-	
+
+	reset_ladder();
```
connorjclark: I don't know exactly where this should be moved to instead, but as it is here, it seems the ladder can go away too soon. for example, if the player warps while on a ladder, then the ladder goes away as the screen transitions.

https://zc-replay-compare-11639869472.surge.sh/

`nargads_trail_crystal_crusades/nargads_trail_crystal_crusades_07_of_24.zplay - 23881`
image
image

the ladder should be present during the first part of the screen transition, but be gone during the second, yeah?

Jambu2024: oh raft not behaving that's interesting.

Jambu2024:
@connorjclark 
their are no changes to the raft so i think we can merge this through now that the checks are passing.

src/zc/hero.cpp
```diff
@@ -25461,36 +25461,40 @@ bool HeroClass::HasHeavyBoots()
...
+
 	// Drawing commands probably shouldn't carry over...
-	if ( !get_qr(qr_SCRIPTDRAWSINWARPS) )
+	if (!get_qr(qr_SCRIPTDRAWSINWARPS))
 		script_drawing_commands.Clear();
-	
+
+	if (replay_version_check(35))
+	{
+		reset_ladder();
```
connorjclark, line 25498:
```suggestion
		// TODO: Would be better to remove a bit later, after fading out the current screen. For now, do this here.
		// https://github.com/ZQuestClassic/ZQuestClassic/pull/1017#discussion_r1828880117
		reset_ladder();
```

connorjclark (approved):

PR #1021, opened 2024-11-09
feat: add enemy types to zinfo

Jambu2024:
dialog works, but upon saving the quest it corrupts the zinfo.


src/dialog/zinfodlg.cpp
```diff
@@ -503,7 +505,66 @@ std::shared_ptr<GUI::Widget> ZInfoDialog::view()
 									}).show();
 							})
 					)
+				)),
+				TabRef(name = "Enemy Types", Rows<3>(
+					vAlign = 0.0,
+					DropDownList(data = list_efamilies,
+						colSpan = 3, fitParent = true,
+						selectedValue = selctr,
```
EmilyV99:
```suggestion
						selectedValue = seletype,
```

EmilyV99 (commented):
Just looking over the code, only found 1 small typo, and no idea why it would be corrupting anything on save...

PR #1022, opened 2024-11-10
refactor: read weapon names from zinfo, not the deprecated module system

Jambu2024:
zcgui listdata lweapons, eweapons, and weapons no longer use zcmodule. they have been removed from their as well.

image
image


Jambu2024:
to explain the change the zcgui listdata lweapons, eweapons, and weapons no longer use zcmodule. they have been removed from their as well.

connorjclark (approved):
🔥

PR #1023, opened 2024-11-11
fix(zc): when charging duration of Charge Ring is 0, use default time instead

Jambu2024:
https://discord.com/channels/876899628556091432/1305279708673343688


connorjclark (approved):
lovely

PR #1026, opened 2024-11-14
refactor: Enemy Editor Attributes Now Use Scroilling Pane

Jambu2024:
image

currently breaks visually when clicking on a entry

image


Jambu2024:
scrubbing this. breaks stuff

PR #1038, opened 2024-12-06
feat: add Misc SFX when the Hero lands on the ground

EmilyV99:


EmilyV99:
bah, failing on metadata, will need to update that

Jambu2024:
does this always play? one would think a different sound would play when landing in water. or on combos with stepsfx

EmilyV99:
> does this always play? one would think a different sound would play when landing in water. or on combos with stepsfx

It plays if you are "standing" after you land- which excludes things like water/pits (swimming/drowning/falling == not standing)

...Water, appears to already play some sort of sound, though I can't find what's causing it, or where it would be set.... so unsure whats up with that.

As for different sound for combos, note the comment `//TODO: per-combo landing SFX? If the combo you're landing on has one, set that, ELSE switch(action)... -Em`. So, something that could be added easily enough in the future; I just didn't feel like dealing with that at the moment.

PR #1040, opened 2024-12-07
fix: bubbles have correct jinx in default quest

Jambu2024:


EmilyV99:
what was the issue?

Jambu2024:
they didnt have the proper effectstrength or rather what to jinx.

EmilyV99:
> they didnt have the proper effectstrength or rather what to jinx.

Like, these? 
image

Those are all set in the default quest already for me....

PR #1041, opened 2024-12-07
feat: (enemy editor) added descriptive labels and dropdowns to walker type attributes

Jambu2024:


src/dialog/enemyeditor.cpp
```diff
@@ -310,6 +310,20 @@ static const GUI::ListData list_deathtype
...
+	{ "Item", e8tITEM },
+	{ "Shield", e8tSHIELD},
+	{ "Sword & Item", e8tSWORD | e8tITEM},
+	{ "Sword & Shield", e8tSWORD | e8tSHIELD},
+	{ "Item & Shield", e8tITEM | e8tSHIELD},
+	{ "All", e8tSWORD | e8tITEM | e8tSHIELD}
+};
+
+#define SW_DROPDOWN 1
+#define SW_TEXTFIELD 0
```
EmilyV99:
```suggestion
#define SW_TEXTFIELD 0
#define SW_DROPDOWN 1
```
OCD.

src/dialog/enemyeditor.cpp
```diff
@@ -648,7 +724,7 @@ void EnemyEditorDialog::loadEnemyType()
 		tf_attributes[q]->setHighBound(999999);
 		if (sw_attributes[q]->getCurrentIndex() != 0) // change this 0 to a constant representing the textfield spot in the switcher
 		{
-			sw_attributes[q]->switchTo(1); // change this 0 to a constant representing the textfield spot in the switcher
+			sw_attributes[q]->switchTo(SW_DROPDOWN); // change this 0 to a constant representing the textfield spot in the switcher
 			tf_attributes[q]->setVal(local_guyref.attributes[q]); //update the value
 		}
```
EmilyV99: Why is this switching to a dropdown when it's already a dropdown? This whole block of code doesn't make sense... Is it supposed to be `switchTo(SW_TEXTFIELD)`? because that would make more sense. Also the `!= 0` should be `!= SW_TEXTFIELD` as well in that case...

EmilyV99 (changes requested):

src/dialog/enemyeditor.cpp
Jambu2024: oh snap good catch.

src/dialog/enemyeditor.cpp
```diff
@@ -1021,149 +1099,279 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::view()
 	case eeWALK:
```
EmilyV99, line 1099: err, this entire `switch(local_guyref.family)` should be part of `loadEnemyType()`, shouldn't it? The code there that's resetting the textfields will override you setting dropdowns here, making all dropdowns vanish whenever you change the enemy type.

EDIT: Except you're reloading the entire dialog from scratch every time you change the enemy type. Which prevents this from being an issue, even though it's kinda bad itself. Beyond the scope of this PR though, so, all good for this.

PR #1043, opened 2024-12-08
feat: attribute 10 now used for hop height for vires/polsvoice

Jambu2024:


src/zc/guys.cpp
```diff
@@ -11506,8 +11506,12 @@ void eStalfos::vire_hop()
 		return;
 		
 	int32_t jump_width = (dmisc9==e9tPOLSVOICE) ? 2 : 1;
-	int32_t jump_height = (dmisc9==e9tPOLSVOICE) ? 27 : 16;
-	
+	int32_t jump_height;
+
+	if ((dmisc10 <= 0 || replay_version_check(0, 37)))
```
EmilyV99: What if someone used dmisc10 in an old quest? It might not be <= 0, and then that old quest would break due to having the wrong values here- AND we wouldn't catch that bug with a replay due to the replay check...

Unsure on if this is ok as-is or not. I'll defer to @connorjclark on this.

Jambu2024: its what we did for ropes but good point

connorjclark:
Can you include in this PR the changes necessary to inform users that this feature exists? At minimum, the label in the enemy editor needs to be set, via `EnemyEditorDialog::loadEnemyType` I think?

src/zc/guys.cpp
connorjclark: I confirmed that none of the replays fail w/o this check, so let's remove the replay bump here.

If we come across a quest which utilized misc10 for vires/pols in this way, we can add a compat rule at that time.

connorjclark (changes requested):
(see my other comments)

connorjclark (approved):

src/zc/guys.cpp
Jambu2024: done

PR #1044, opened 2024-12-09
refactor: enemy editor

Jambu2024:
@EmilyV99

### DIFF
--- a/src/dialog/enemyeditor.cpp
+++ b/src/dialog/enemyeditor.cpp
@@ -433,6 +433,7 @@ void EnemyEditorDialog::loadAttributes()
 		ddl_attributes[0]->setOnSelectFunc([&](int32_t val) //summoning
 			{
 				local_guyref.attributes[0] = val;
+				tf_attributes[0]->setVal(val);
 				switch (val)
 				{
 				case e1tSUMMON:
@@ -453,6 +454,7 @@ void EnemyEditorDialog::loadAttributes()
 		ddl_attributes[1]->setOnSelectFunc([&](int32_t val) //deatheffect
 			{
 				local_guyref.attributes[1] = val;
+				tf_attributes[1]->setVal(val);
 				if (local_guyref.attributes[0] == e1tSUMMON)
 				{
 					return;
@@ -488,6 +490,7 @@ void EnemyEditorDialog::loadAttributes()
 		ddl_attributes[6]->setOnSelectFunc([&](int32_t val) //toucheffect
 			{
 				local_guyref.attributes[6] = val;
+				tf_attributes[6]->setVal(val);
 				switch (val)
 				{
 				default:
@@ -524,6 +527,7 @@ void EnemyEditorDialog::loadAttributes()
 			});
 		ddl_attributes[8]->setOnSelectFunc([&](int32_t val) { //movestyle
 			local_guyref.attributes[8] = val;
+			tf_attributes[8]->setVal(val);
 			switch (val)
 			{
 			default:
@@ -1362,7 +1366,8 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::view()
 				onValChangedFunc = [&, q](GUI::TextField::type, std::string_view, int32_t val)
 				{
 					local_guyref.attributes[q] = val;
-					loadAttributes();
+					ddl_attributes[q]->setSelectedValue(val);
+					loadAttributes(); //shouldn't be need but just in case
 				}),
 			ddl_attributes[q] = DropDownList(
 				width = 300_px,
@@ -1373,6 +1378,8 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::view()
 				onSelectFunc = [&, q](int32_t val)
 				{
 					local_guyref.attributes[q] = val;
+					tf_attributes[q]->setVal(val);
+					loadAttributes();
 				})
 		));
 	}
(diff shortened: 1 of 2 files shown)


src/dialog/enemyeditor.cpp
```diff
@@ -1985,6 +1984,7 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::view()
 			)
 		)
 	);
+	loadEnemyType();
```
EmilyV99:
```suggestion
```
`refreshScript()` already calls `loadEnemyType()`, so now you're calling it all twice.

src/dialog/enemyeditor.cpp
```diff
@@ -1026,355 +1308,72 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::WeaponScriptField(int index)
...
-			});
-		ddl_attributes[6]->setOnSelectFunc([&](int32_t val) //toucheffect
+
+std::shared_ptr<GUI::Widget> EnemyEditorDialog::view()
+{
+	using namespace GUI::Builder;
+	using namespace GUI::Props;
+	using namespace GUI::Key;
+
+	//MACRO DEFINES GO HERE LOCALLY INSTEAD OF IN GLOBAL SCOPE
```
EmilyV99:
```suggestion
```
Misleading comment, macros aren't scoped

src/dialog/enemyeditor.cpp
```diff
@@ -2015,14 +2015,14 @@ bool EnemyEditorDialog::handleMessage(const GUI::DialogMessage<message>& msg)
...
+		//rerun_dlg = true;
+		return false;
 	}
 	case message::SETALLDEFENSE:
 	{
 		for (int q = 0; q < edefLAST255; ++q)
 			local_guyref.defense[q] = local_guyref.defense[0];
-		loadEnemyType();
+		//loadEnemyType();
 		rerun_dlg = true;
```
EmilyV99, line 2033: This section should probably directly set the value of the defense dropdowns, instead of rerunning the dialog.

EmilyV99 (changes requested):

src/dialog/enemyeditor.cpp
```diff
@@ -1020,361 +1306,75 @@ std::shared_ptr<GUI::Widget> EnemyEditorDialog::WeaponScriptField(int index)
...
-		sw_attributes[9]->switchTo(SW_DROPDOWN);
-		break;
-	case eeNONE: //(Boss Death Triggers)
-		ddl_attributes[9]->setSelectedValue(local_guyref.attributes[9]);
-		ddl_attributes[9]->setListData(list_yesnomisc);
-		sw_attributes[9]->switchTo(SW_DROPDOWN);
-		break;
-	default:
-		break;
+					local_guyref.attributes[q] = val;
```
EmilyV99, line 1380: missing `loadAttributes()` here

src/dialog/enemyeditor.cpp
```diff
@@ -395,6 +395,291 @@ void EnemyEditorDialog::refreshScript()
 	}
 }
 
+void EnemyEditorDialog::loadAttributes()
+{
+	switch (local_guyref.family)
+	{
+	case eeWALK:
+		ddl_attributes[0]->setSelectedValue(local_guyref.attributes[0]);
```
EmilyV99, line 403: You're updating the selected value of the dropdown, but not the associated textfield?

EmilyV99 (changes requested):

src/dialog/enemyeditor.cpp
Jambu2024: i tried that, it doesn't update the dialog. leave it alone.

EmilyV99: what did you try that didn't work?
Maybe try adding `pendDraw()` after it?

src/dialog/enemyeditor.cpp
Jambu2024: am i suppose to?

EmilyV99: If you edit the textfield, it changes the dropdown; so why would the other way not happen?

What would happen here is, the enemy's attribute would be whatever the dropdown set it to, but it would show a different value in the textfield to the user- if they hit OK then, it would save the value from the dropdown, which they can't see. That definitely would be a bug.

Jambu2024: its saves the values from the local_guyref the textfields and ddls are just widgets to display data nothing more

EmilyV99: > its saves the values from the local_guyref the textfields and ddls are just widgets to display data nothing more

Yes. Which means you could be looking at a textfield that says `4`, hit `OK`, and have it save `10` as the value instead of `4`, because the textfield is showing you, the user, the wrong value.

Jambu2024: i have never seen this happen emily/ until now see chat.

Jambu2024: actually seems impossible since both write to local_guyref

src/dialog/enemyeditor.cpp
Jambu2024: wait i dont even have shared pointers for the defense dropdowns defined in the class sign.

EmilyV99: yeah, you'd need to add those then

src/dialog/enemyeditor.cpp
Jambu2024: should be fixed now

Jambu2024:
do it yourself

PR #1045, opened 2024-12-10
Refactor enemyeditor

EmilyV99:
takes over from #1044


Jambu2024:
seems we will have to revert the bad commit first, and then refactor the enemy editor off that. then reimplement the changes the right way last.

PR #1051, opened 2024-12-17
fix(zq): remove hidden combo selection mode toggled with spacebar

EmilyV99:
The setting 'Allow Non-Column Combo View' allows knowing users to opt-in. Additionally, the keybind is no longer bound by default- though this only affects the main view, not the combo pages view.

This notably becomes an issue when a user has the setting toggled in the main view, where there is ***no visual indicator that it is toggled on***, and it completely changes how combos are placed when using a brush size >1x1. Deathrider in his most recent dev stream ran into an issue due to this, and, not knowing any better way to fix it, simply restarted ZQ entirely (as the `combo_cols` setting is not saved on exit, so that happens to revert it).

Discussion: We could potentially also just trim it entirely, but I'm unsure if it has its' particular uses that some users (or some old tilesets?) might miss having it available for?


EmilyV99:
Would squash as "fix(zq): Remove `combo_cols` setting when hitting spacebar".
Changing the combo "pages" to use the same system as the autocombo/pool/alias "pages" would be a good next step, as that could offer the multiple modes of selection in the pages inherently.

connorjclark:
related discussion [here](https://discord.com/channels/876899628556091432/1116933752967405578)

seems great to me. my commit suggestion, feel free to modify

```
fix(zq): remove hidden combo selection mode toggled with spacebar

Pressing space changed how combo selection worked in a way that often confused
users. There was also no visible indication of this, so it was extra bad. This
mode was perhaps useful for something in the combo selector dialog, but outside
that it was seemingly not useful at all - and is now removed.
```

connorjclark (approved):

PR #1053, opened 2024-12-26
feat: subscreen counters can have separate color when "full"

EmilyV99:


resources/include/bindings/subscreenwidget.zh
```diff
@@ -391,6 +391,14 @@ class subscreenwidget {
...
 	// @zasm_var SUBWIDGTY_COLOR_TXT
 	internal int ColorText;
+	
+	// The widget's secondary color index used for Text
+	//
+	// - [SUBWIDG_COUNTER]
+	// - [SUBWIDG_BTNCOUNTER]
+	//
+	// @zasm_var SUBWIDGTY_COLOR_TXT2
+	internal int ColorText2;
```
connorjclark: rename "SecondaryColorText" (or w/e)?

is it not always what's used "when full"? if it is, why not "FullColor"?

EmilyV99: Could re-use if another widget has a "second" or "alternate" text color, which might not be a "FullColor"- unless you think adding separate zasm vars that only apply for like one thing like that is fine (i.e. name this FullColor, and then in the future for some other widget name a new var specifically instead of reusing the same var)

connorjclark: Separate, specific names would be easier to document and read/understand. If we end up with another case where we need a new color variable, we could even re-use the ZASM variable if there's no overlap in usage (so multiple variable names all as an alias to `SUBWIDGTY_COLOR_TXT2` could be fine).

PR #1054, opened 2024-12-26
refactor(zq): Move combo pages to new page gui system

EmilyV99:
Notable TODOS for things that need to be brought from the old dialog:
- 'Insert' and 'Remove' functionality
- 'Overwrite Protection' warning you when you're pasting over something in use
- 'Move' option to move combos properly
- Finally, clear out the old dialog code once it's no longer used.

### DIFF
--- a/src/dialog/externs.h
+++ b/src/dialog/externs.h
@@ -5,6 +5,7 @@
 #include "base/containers.h"
 struct mapscr;
 
+void call_cmb_pages(optional<int> val = nullopt);
 void call_cpool_pages(optional<int> val = nullopt);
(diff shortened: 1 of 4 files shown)


Jambu2024:
TODO:
- Box Selector for Combo Page. It currently only allows sequences to be selected which isnt very helpful.
- Add SelComboSwatchSet for above pages.

EmilyV99:
> TODO:
> * Box Selector for Combo Page. It currently only allows sequences to be selected which isnt very helpful.

that's already part of the existing system the other things use, so, that would be part of the title of the PR

PR #1056, opened 2025-01-03
Green square fix tool + AlertFuncDialog refactor

EmilyV99:


src/dialog/alertfunc.cpp
```diff
@@ -5,34 +5,51 @@
...
+	std::initializer_list<std::function<bool()>> buttonProcs)
 {
-	initButtons(buttonNames, buttonProcs, numButtons, focused_button);
+	initButtons(buttonNames, buttonProcs, focused_button);
+	return *this;
 }
-
-AlertFuncDialog::AlertFuncDialog(string const& title, vector<string> const& lines,
-	string info, uint32_t numButtons, uint32_t focused_button,
+AlertFuncDialog& AlertFuncDialog::add_buttons(uint32_t focused_button,
```
connorjclark, line 23: for this variant, would you consider removing 'focused_button' and just having the initial value of `chosen` be used for selecting the initial focus value? (would still set value to -1 if user closes the dialog with ESC/the X button)

connorjclark (approved):

src/dialog/alertfunc.cpp
EmilyV99: On trying to implement this, it felt kinda bad/inconsistent IMO. Manually setting the chosen to -1 does seem more consistent though, so pushed that.

PR #1058, opened 2025-01-03
fix(zc): Custom Bush/Slash combo sprites not spawning centered on the combo

EmilyV99:
Likely need to update some replays still; but only things using a custom `comboSprite` are expected to change, standard bush clippings and such should remain as usual.


EmilyV99:
...or no replay failures on this. Guess we don't have coverage of custom sprites used on slash combos anywhere. Makes sense, since, who would use them with this bug existing?

PR #1060, opened 2025-01-08
Updatebindings

EmilyV99:
- Updates a bunch of `arg1`, `arg2` args to have names
- renames a couple args to be more descriptive (ex. `rangle` -> `rot_degress`)
- Fixes a bad binding (3.0 only) for DrawCombo in Screen and Bitmap
- Fixes bad bindings (still need to be fixed in 2.55!) for bitmap special layer drawing functions


resources/include/bindings/global.zh
```diff
@@ -908,56 +908,64 @@ internal int Integer(int n);
 // @zasm
 //   POP D0
 //   SETR D2 SCRIPTRAMD
-internal int GetScriptRAM(int arg1);
+internal int GetScriptRAM(int idx);
```
connorjclark: `index` - lets make sure it's clear to everyone what these mean

resources/include/bindings/hero.zh
```diff
@@ -397,6 +406,12 @@ class hero {
 	// @zasm_var HEROHAMMERSTATE
 	internal int HammerState;
 
+	// When the engine lifts an object, it will overwrite these flags.
+	// - LIFTFL_DISABLE_SHIELD - If true, the shield is disabled when lifting the same way it is when swinging a weapon.
```
connorjclark: wrap symbols in square brackets - `[LIFTFL_DISABLE_SHIELD]` - so that it gets linkified in vs code tooltips and in the docs website

resources/include/bindings/hero.zh
```diff
@@ -803,14 +818,19 @@ class hero {
...
 
+	// Returns a pointer to the held weapon, and causes the player to let go of it.
+	//
 	// @zasm
 	//   HEROLIFTRELEASE
 	internal lweapon ReleaseLiftWeapon();
-
+	
+	// Causes the player to lift the specified weapon in the same way they lift combos with the Lift Glove.
+	// If an lweapon script lifts itself, this also functions as a Waitframe().
```
connorjclark:
```suggestion
	// If an [lweapon] script lifts itself, this also functions as [Waitframe].
```

etc.

resources/include/bindings/global.zh
```diff
@@ -908,56 +908,64 @@ internal int Integer(int n);
...
 
 // @deprecated
 // @zasm
 //   POP D3
 //   POP D0
 //   SETR GLOBALRAMD D3
-internal void SetGlobalRAM(int arg1, int arg2);
+internal void SetGlobalRAM(int idx, int val);
 
+// Unimplemented, never did anything
```
connorjclark: I confirmed never used in zscript database.

let's delete all these unimplemented ones to kill some noise inside the docs.

can also mark the commands as deleted in code... i've done that with a number of things recently like this:

```c++
// src/zasm/defines.h
DELETED_SETCOLORB,
DELETED_SETDEPTHB,
DELETED_GETCOLORB,
DELETED_GETDEPTHB,
```

(and removing their usages) - since we know it isn't used, but don't necessarily want to reuse the command numbers.

connorjclark:
merged as https://github.com/ZQuestClassic/ZQuestClassic/commit/b34772036c12cfd3fc3a42908f0a838d5189bbfc

PR #1070, opened 2025-01-25
fix(zc): progressive items related crash

EmilyV99:
having a prog item grant itself or other prog items now prints an error message rather than crashing the player.

Not yet fixed ( @connorjclark maybe you can finish this, I need to rest):

A prog item that grants an item bundle that grants a prog item. (Or item bundles that grant themselves). These need a fix in `hero.cpp` in `getitem`.

~~Should abstract this as `vector<int32_t> get_related_items(int32_t id)`, which:~~
- ~~For any item bundle, it recursively calls itself for every item inside. It returns the combination of all the results, NOT including the item ID of the item bundle itself.~~
- ~~For any progressive item, calls `get_progressive_item()` and recursively calls itself with the result.~~
- ~~For any other item, it returns just that item ID.~~
- ~~Handles anything that would cause an infinite loop with an error message (ex. a bundle that grants itself, or grants a progressive item that then grants the bundle)~~
  - ~~NOTE: A bundle that grants multiple copies of another bundle is PERFECTLY VALID, and would form no infinite loop- such a case must remain valid.~~

~~This results in a vector of all the items that need to be granted to the player. The recursion would need a similar handling to the one I already coded for progressive items, to prevent re-visiting (though has less use cases to handle, so a simple set for visited numbers might be enough). This would be called in `getitem` for `itype_itembundle` specifically, and importantly, could also be given ***script access*** (as `void Game->GetRelatedItems(int[] buf, int id);`?) as a way to solve~~ [this issue](https://discord.com/channels/876899628556091432/1332392366551601203/1332461053249847483)

Except, abstracting it in that way won't handle the case of multiple progressive items properly, because if a bundle were to contain two of the same progressive item, getting the first one would cause the second to progress- or even a bundle of a non-progressive item that AFFECTS a progressive item, like putting a small quiver in with a 'progressive quiver'. So, doing that is probably "wrong". Just a more simple visited set in `getitem` might do better? That won't solve the script access issue, though, I don't actually see a good solve for the script access issue... unless we wanted to write a whole separate calculation just to handle that....


Jambu2024:
ack stupid merged commit

Jambu2024:
should be read to go after we remove the merge commit.

EmilyV99:
TODO: be sure that an item bundle that grants *itself*, as well as an item bundle A granting an item bundle B that grants item bundle A, do not crash.

PR #1072, opened 2025-01-29
fix(zc): 'slashable' combos (tall grass, etc) wrongly 'cutting' via triggers tab

EmilyV99:
Due to some old half-finished code, slashables were responding to all weapons set in their combo triggers tab, even with `->ComboType Effects` not checked. This behavior has been fixed, now requiring `->ComboType Effects` to trigger.

### DIFF
--- a/src/core/qst_combos.cpp
+++ b/src/core/qst_combos.cpp
@@ -52,7 +52,7 @@ int32_t init_combo_classes()
     return 0;
 }
 
-int32_t readcombos_old(word section_version, PACKFILE *f, zquestheader *, word version, word build, word start_combo, word max_combos)
+int32_t readcombos_old(word section_version, PACKFILE *f, zquestheader *Header, word version, word build, word start_combo, word max_combos)
 {
 	bool should_skip = legacy_skip_flags && get_bit(legacy_skip_flags, skip_combos);
 	byte tempbyte;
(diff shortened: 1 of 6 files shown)


Jambu2024:
seems this can be merged emily. you need to resolve the conflicts first tho... no it cannot, i am blind... my bad.

Jambu2024 (approved):

EmilyV99:
> seems this can be merged emily. you need to resolve the conflicts first tho

It was failing replays though? so uh, no it can't

Jambu2024 (dismissed):
So the reason the replays are failing is do to slash triggers no displaying slashing animation anymore in those few quest when a trigger is replayed. Easy Fix, just add a replay check no quest rule needed. Once this and all other checks pass it should be fine

FAIL: test_save (test_zeditor.TestZEditor.test_save) [ss_jenny.zplay]

src/core/qst_combos.cpp
```diff
@@ -474,6 +474,11 @@ int32_t readcombos_old(word section_version, PACKFILE *f, zquestheader *, word v
 			temp_combo.triggers.clear();
 		
 		update_combo(temp_combo, section_version);
+		if(section_version < 67 && !(Header && Header->version_major == 2 && Header->version_minor == 55 && Header->version_patch >= 14))
```
connorjclark, line 477: `if (section_version < 67 && Header->compareVer(2, 55, 14) < 0)` should work too i think?

1. This method handles version checks comprehensively, while the current code may not be forward-compatabile (altho it may not matter IIRC in combination with the section version check).
2. `Header` is never null

connorjclark (approved):
Uploaded replays all pass 👍

src/core/qst_combos.cpp
EmilyV99: . . . you are technically correct. It did make more sense to me though to have the two spots that are checking this both be identical, and in the other spot, that version check would fail to catch 3.0 versions with the issue, and header specifically can be null there. Having the check be identical feels good for consistency.

PR #1078, opened 2025-04-06
New drawing stuff

EmilyV99:
- sprite Z-sort and optional Y-sort
- z-thresholds for each layer, allowing sprites to "jump through" layers as their Z increases
- weapon animation code is no longer done during weapon drawing (also fixes weapon shadow draw timing)
- generally cleaner code
- some minor graphical fixes


resources/include/bindings/qrs.zh
```diff
@@ -395,6 +395,29 @@ enum QR
 	qr_SPOTLIGHT_IGNR_SOLIDOBJ,	// Compatibility (>>>) Light Beams Ignore Solid Objects
 	qr_BROKEN_LIGHTBEAM_HITBOX,	// Compatibility (>>>) Broken light beam hitbox
 	qr_BROKEN_SWORD_SPIN_TRIGGERS,	// Compatibility (>>>) Broken Sword Triggers (spin-related)
+	
+	// TODO document these
+	qr_SCRIPT_CONTHP_IS_HEARTS,
```
connorjclark: this comment would show up in the docs, remove?

src/zc/maps.cpp
```diff
@@ -5091,14 +5022,602 @@ void draw_screen(bool showhero, bool runGeneric)
...
+		do_primitives(framebuf, SPLAYER_DARKROOM_OVER);
 	}
 	
+	do_ffc_layer(framebuf, 7);
 	draw_msgstr(7);
 	
 	set_clip_rect(scrollbuf, 0, 0, scrollbuf->w, scrollbuf->h);
 	if(runGeneric) FFCore.runGenericPassiveEngine(SCR_TIMING_POST_DRAW);
 }
+static void draw_sprites_new(BITMAP* dest, set<sprite*, SpriteSorter>& sprite_set, set<sprite*, SpriteSorter>::iterator& it, word upto_z)
```
connorjclark: rename to just `draw_sprites`?

src/zc/maps.cpp
```diff
@@ -5091,14 +5022,602 @@ void draw_screen(bool showhero, bool runGeneric)
...
+{
+	clear_info_bmp();
+	if((GameFlags & (GAMEFLAG_SCRIPTMENU_ACTIVE|GAMEFLAG_F6SCRIPT_ACTIVE))!=0)
+	{
+		FFCore.doScriptMenuDraws();
+		return;
+	}
+	if(get_qr(qr_CLASSIC_DRAWING_ORDER))
+	{
+		draw_screen_old(showhero, runGeneric);
```
connorjclark: this is an enormous potential source of bugs, given its a total separate branch of code. None of our 100s+ hours of replays will hit anything new in this function, so its difficult to know if its correct. And also, it's doubling our footprint for code to maintain: any new feature that adds to drawing code will have to touch both - and if we only have replays for one branch, we're gonna make mistakes.

Is there another way to integrate the new behavior than doubling up our code here?

resources/include/bindings/qrs.zh
EmilyV99: probably better to add comments for each of these instead, then

src/zc/maps.cpp
```diff
@@ -5053,22 +4976,21 @@ void draw_screen(bool showhero, bool runGeneric)
 	// Draw some text on framebuf
 	
 	set_clip_rect(framebuf,0,0,256,232);
+	draw_msgstr(6);
```
connorjclark: i suppose this draw_msgstr can't be closer to the other layer 6 calls (do_ffc_layer/particles.draw) due to the darkness effect?

resources/include/bindings/ffc.zh
```diff
@@ -87,7 +92,9 @@ class ffc {
 	//
 	// @zasm_var YD2
 	internal int Ay;
-
+	
+	// Flags for the FFC. Uses [FFCFlag].
+	//
```
connorjclark: `// @index [FFCFlag]` (Ex: see `Screen->State`)

src/zc/maps.cpp
EmilyV99: aye, that string layer is supposed to be above that darkness

resources/include/bindings/ffc.zh
EmilyV99:
```suggestion
	// Flags for the FFC.
	//
	// @index [FFCFlag]
```

resources/include/bindings/qrs.zh
connorjclark: this might be a little overkill, but it would be somewhat simple to make a `-dev-qrs-script` debug command that spits out all this for us. For all we know there's an off-by-one error here somewhere 👀

EmilyV99: > this might be a little overkill, but it would be somewhat simple to make a `-dev-qrs-script` debug command that spits out all this for us. For all we know there's an off-by-one error here somewhere 👀

Well, I added these here because I was using them in my test script, so can confirm no off-by-one error or my test script wouldn't have worked. But aye, more automation is always good.

EmilyV99: > this might be a little overkill, but it would be somewhat simple to make a `-dev-qrs-script` debug command that spits out all this for us. For all we know there's an off-by-one error here somewhere 👀

Removed the update entirely for now, esp since this isn't being merged soon- an automatic updater like that would be very nice to add at some point, and I definitely think is the right solution in the long run. Maybe we could even make the doc comments be the full `i` button text from `quest_rules.cpp`?

src/zc/maps.cpp
EmilyV99: K, reduced the double-up as much as possible. Anything that has to do with drawing enemies, items, weapons, etc still is split, because those are entirely changed, so can't really be reduced as it isn't duplicate.

src/dialog/ffc_editor.cpp
```diff
@@ -385,7 +385,7 @@ std::shared_ptr<GUI::Widget> FFCDialog::view()
 							}
 						),
 						//
-						NOSWAP_FIELD("Layer:", ffc.layer, 0, 7, 
+						NOSWAP_FIELD("Layer:", ffc.layer, -7, 7,
 							"The layer (0-7) the ffc will be drawn on. Ignored if 'Draw Over' is checked."
 							" Will draw above the combos AND script draws on the specified layer.")
```
connorjclark: should update description if negative values are ok (and maybe explain what it means to be negative / they aren't real combo layers)

src/sprite.h
```diff
@@ -338,6 +346,19 @@ class breakable : public sprite
...
 };
 
+class tempsprite_shadow : public sprite
+{
+public:
+	tempsprite_shadow(sprite* parent);
+	int32_t getUID() override
+	{
+		return parent ? parent->getUID() : 0;
+	}
```
connorjclark: would like to avoid making this virtual if we can (for performance reasons) and logic reasons (it's confusing that two sprites can have the same uid – that would confuse sprite lists).

is this just to influence the behavior of `SpriteSorter::operator()`? If so, that method is already checking if a type is this shadow subclass, so can't we grab the paren't UID there explicitly?

src/zc/scripting/types/game.cpp
```diff
@@ -1315,6 +1315,13 @@ static ArrayRegistrar GAMEGRAVITY_registrar(GAMEGRAVITY, []{
 	return &impl;
 }());
 
+static ArrayRegistrar GAMELAYERZTHRESHOLDS_registrar(GAMELAYERZTHRESHOLDS, []{
+	static ScriptingArray_GlobalCArray impl(zinit.sprite_z_thresholds, comptime_array_size(zinit.sprite_z_thresholds));
+	impl.setDefaultValue(-10000);
```
connorjclark: I'd like to default to 0 for new arrays unless there's a strong reason not to. Is there here?

connorjclark: Note: once the negative-indices feature is supported for these internal arrays, the "default value" will never be used... at least, not for global c-arrays. It'll just wrap the index so no need for a default.

I guess a default might still be used for the `Object` getters when the object can't be resolved. But I might take that opportunity to make even that return 0 too. In any case, for most (or all) usages of `default value` there will soon be a QR that mostly (if not totally) disables its usage.

PR #1080, opened 2025-04-09
feat: add "screen load" combo trigger

arceusplayer11:
Also fixed some dark room bugs.

### DIFF
--- a/src/zc/maps.cpp
+++ b/src/zc/maps.cpp
@@ -1172,7 +1172,7 @@ static void apply_state_changes_to_screen(mapscr& scr, int32_t map, int32_t scre
 	
 	for_every_combo_in_screen(screen_handles, [&](const auto& handle) {
 		auto& cmb = handle.combo();	
-		if (cmb.triggerflags[4] & combotriggerSCREENENTRY)
+		if (cmb.triggerflags[4] & combotriggerSCREENLOAD)
 		{
 			do_trigger_combo(handle);
 		}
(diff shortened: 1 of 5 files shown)


src/zc/maps.cpp
```diff
@@ -1163,10 +1163,19 @@ static void apply_state_changes_to_screen(mapscr& scr, int32_t map, int32_t scre
...
 
 	int mi = mapind(map, screen);
 	clear_xdoors_mi(screen_handles, mi);
 	clear_xstatecombos_mi(screen_handles, mi);
+	
+	for_every_combo_in_screen(screen_handles, [&](const auto& handle) {
+		auto& cmb = handle.combo();	
+		if (cmb.triggerflags[4] & combotriggerSCREENENTRY)
+		{
+			do_trigger_combo(handle);
```
connorjclark, line 1177: This function is currently not safe to call on external regions. I'll need to fix it before this code lands.

also... MAPCOMBO3 will call this function (potentially many times every frame), so I think this could potentially do stuff like play an sfx tons of times every frame.

arceusplayer11: Ah, shoot. I'm not super used to this system, haha.

src/zc/maps.cpp
```diff
@@ -6346,6 +6361,14 @@ void loadscr_old(int32_t destdmap, int32_t screen,int32_t ldir,bool overlay)
...
 	clear_xdoors(screen_handles, true);
 	clear_xstatecombos(screen_handles, true);
+	
+	for_every_combo_in_screen(screen_handles, [&](const auto& handle) {
+		auto& cmb = handle.combo();	
+		if (cmb.triggerflags[4] & combotriggerSCREENENTRY)
+		{
+			do_trigger_combo(handle);
+		}
+	});
```
connorjclark, line 6309: Instead of calling it in loadscr (so it would trigger as soon as a region loads, even if the player is nowhere near it), do you think it'd be more useful to call it only when the screen first is close to the player/in view? That's when enemies load right now.

if not, that's fine - just change the text description to "Trigger on entry (when region loads)"

arceusplayer11: I think triggering when the region loads is more useful IMO; the intended use case here is to have a combo turn on or off darkness when entering a screen.

PR #1088, opened 2025-05-23
Subscreen conditionals

EmilyV99:


src/zc/ffscript.cpp
```diff
@@ -11727,6 +11727,70 @@ int32_t get_register(int32_t arg)
...
 		}
+		case SUBWIDGREQOWNITEMS:
+		{
+			if(SubscrWidget* widg = checkSubWidg(ri->subwidgref))
+			{
+				size_t indx = ri->d[rINDEX]/10000;
+				if(indx >= MAXITEMS)
+				{
+					Z_scripterrlog("Bad index '%d' to array "
+						"'subscreenwidget->RequiredOwnedItems[%d]'\n", indx, MAXITEMS);
```
connorjclark, line 11738: prefer `scripting_log_error_with_context` now (still need to migrate many/most remaining Z_scripterrlog)

note: it uses `fmt`, the trailing newline is not needed, and the name of the method should be elided (as `scripting_log_error_with_context` will add it)

src/zasm/table.cpp
```diff
@@ -1166,6 +1166,8 @@ static constexpr script_command command_list[]=
 	{ "BMPCOMBOBLIT", BMPCOMBOBLIT, 0, {}, 0, 0 },
 	{ "LOAD_FFC", LOAD_FFC, 1, {REG_RW}, 0, 0},
 	{ "REPAIRSHIELD", REPAIRSHIELD, 1, { REG_R }, 0, 0 },
+	{ "SUBWIDG_CHECK_CONDITIONS", SUBWIDG_CHECK_CONDITIONS, 1, { REG_R }, 0, 0 },
+	{ "SUBWIDG_CHECK_VISIBLE", SUBWIDG_CHECK_VISIBLE, 1, { REG_R }, 0, 0 },
```
connorjclark: shouldn't these both be REG_W?

src/zasm/table.cpp
```diff
@@ -3484,6 +3495,8 @@ std::initializer_list<int> get_register_dependencies(int reg)
 		case SUBWIDGBTNPG:
 		case SUBWIDGBTNPRESS:
 		case SUBWIDGFLAG:
+		case SUBWIDGREQOWNITEMS:
+		case SUBWIDGREQUNOWNITEMS:
 		case SUBWIDGGENFLAG:
```
connorjclark, line 3500: (just a note to myself...)

i'm very glad you're catching these need to be added. it's just for the player zasm optimizer. I want to figure out a more automated/declarative way to handle this; same for the REG_W/R mentioned in my other comment. Would like it to somehow be impossible for these things to be wrong.

resources/include/bindings/subscreendata.zh
```diff
@@ -77,11 +77,14 @@ enum //Visibility Flags
 	SUBVISIB_MAX
 };
 
-enum //Widget genflags
-{
-	SUBW_GENFL_SELECTABLE,
-	SUBW_GENFL_PGMODE_NOWRAP,
-	SUBW_GENFL_SELECTOR_OVERRIDE,
+enum SubscreenWidgetGenFlag //Widget genflags
```
connorjclark: This trailing comment (on the enum declaration; the others are fine) doesn't get picked up by the doc comment parser.

(though I would just remove it; since it's just repeating the name)

src/zasm/table.cpp
EmilyV99: ...I could have sworn I set those to REG_W, maybe I forgot to save the file with that change or something? Nice catch lol, thanks

src/zasm/table.cpp
EmilyV99: I just ctrl+f for a recent similar opcode/register and check all the instances when I go to add a new one, so yeah I tend to catch new spots like this.

More automation is always good.

src/zc/ffscript.cpp
```diff
@@ -11727,6 +11727,70 @@ int32_t get_register(int32_t arg)
...
+		}
+		case SUBWIDGREQUNOWNITEMS:
+		{
+			if(SubscrWidget* widg = checkSubWidg(ri->subwidgref))
+			{
+				size_t indx = ri->d[rINDEX]/10000;
+				if(indx >= MAXITEMS)
+				{
+					scripting_log_error_with_context("Bad index '{}' to array "
+						"'subscreenwidget->RequiredUnwnedItems[{}]'", indx, MAXITEMS);
```
connorjclark, line 11752: typo here.. this is why I suggest we don't try to show the names in the error message (just defer to the context string given by `scripting_log_error_with_context`).

Also the `'subscreenwidget->RequiredUnwnedItems[MAXITEMS]'` doesn't make sense to me... but I see other error messages are doing that to indicate the size of the array.

I'll update this to use a common index bounds helper and look into doing this more consistently across all array accesses later.

PR #1093, opened 2025-06-01
fix(zc): some drawing commands ignore the `Toggle Transparency` combo animation flag

EmilyV99:
No compat qr added, only some scripts would be affected (only 2.55+ to have the Toggle Transparency), and these most likely would be desired changes. No replays appear to regress from this.

Can probably be backported to 2.55 as well?


Jambu2024 (approved):
Yeah this should be relatively easy to backport.

connorjclark (approved):

PR #1097, opened 2025-06-02
feat: multiple triggers per combo

EmilyV99:
DONE:
- Each combo can have anywhere between 0-255 combo triggers. Only the used triggers take up space (in memory or qst).
- The triggers are checked in order, so if the same combo has multiple triggers with the same cause, they will run in the specified order.
  - If in such a case, the combo is changed while there are still triggers remaining, the extra triggers will NOT be executed- the execution halts when the combo is altered.
- Editable string Label for each trigger, used in the display name in the combo editor list
- Script access. The old access should all still work, accessing the combo's *first trigger*. Access for new triggers uses the new `combotrigger` datatype.
  - `Screen->TriggerCombo()` can specify trigger index
  - the 'Only Gen Triggers' flag is now part of combos instead of triggers, so has separate access.
  - command to get a specific trigger by it's label (where invalid names safely return a null value)

TODO:
- nothing left


Jambu2024 (changes requested):
I honestly prefer buttons be the same width, other devs how you feel?

image

connorjclark (approved):

Jambu2024 (approved):

PR #1098, opened 2025-06-05
refactor(zq)! Updated door selector to zcgui

Jambu2024:


Jambu2024:
image

Jambu2024:
Didnt catch this bug before I pushed.
For some reason message::OK is not saving the doors when those are changed (it saves the door combo set).

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
 using std::string;
 using std::to_string;
 
 extern bool saved;
 
-static bool edited = false;
+void set_dlg_active(bool active)
+{
+	pause_dlg_tint(active);
+	zq_set_screen_never_freeze(active);
```
EmilyV99, line 18: These prevent the background from tinting, and make sure it keeps animating/changing. Need to be sure to call to set these properly, or it can leak into other dialogs!

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
+}
+#define DOORSEL_INFOBTN(inf) \
+Button(forceFitH = true, text = "?", \
+	onClick = message::REFR_INFO, \
+	onPressFunc = [=]() \
+	{ \
+		set_dlg_active(false); \
+		displayinfo("Info",inf); \
+		set_dlg_active(true); \
+	})
```
EmilyV99, line 28: A custom version of `INFOBTN`, that calls the `set_dlg_active` function before/after the info popup. Anything you do here that pops up another dialog would need to do something like this.
...Yeah, I didn't actually USE this- but I wrote it to be an example of what would need to be done.

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
 
 bool call_doorseldialog()
 {
-	edited = false;
-	DoorSelDialog().show();
+	set_dlg_active(true);
+	Map.StartListCommand();
+	DoorSelDialog(*Map.CurrScr()).show();
+	set_dlg_active(false);
```
EmilyV99, line 35: Set the dialog active state before and after the dialog.
Also, `StartListCommand` begins a "list" of commands. These "commands" are what are undone/redone via undo/redo. An entire "list" is undone at once when a list is used- all the changes made in the dialog count as a single undo, so a single list.

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
-DoorSelDialog::DoorSelDialog() :
+DoorSelDialog::DoorSelDialog(mapscr const& ref) :
+	screen(ref),
 	list_doortypes(GUI::ZCListData::doortypes()),
 	list_doorsets(GUI::ZCListData::doorsets())
-{}
+{
+	door_combo_set = (old_door_combo_set = screen.door_combo_set);
+	for (int q = 0; q < 4; q++)
+		doors[q] = screen.door[q];;
```
EmilyV99, line 46: It's cleaner to store things like this in the constructor rather than `view` - as long as you don't need them to run again when you `rerun_dlg`, that is.

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
+}
+
+void DoorSelDialog::set_doors()
+{
+	//Recover the combos that were there previously!
+	Map.RevokeListCommand();
+	Map.StartListCommand();
```
EmilyV99, line 53: Revoking the list command "undoes" anything inside it, so, this cancels all changes and starts a fresh list.

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
+	Map.DoSetDCSCommand(door_combo_set);
+	for (int q = 0; q < 4; q++)
+		Map.DoPutDoorCommand(q,doors[q],door_combo_set!=old_door_combo_set);
```
EmilyV99, line 56: Setting these commands is what actually changes the screen, and puts the changes in the undo list.

src/dialog/door_sel.cpp
```diff
@@ -3,25 +3,59 @@
...
+	refresh(rMAP | rNOCURSOR);
```
EmilyV99, line 57: This draws the screen again, to update the changed combos in the background.

src/dialog/door_sel.cpp
```diff
@@ -31,11 +65,11 @@ std::shared_ptr<GUI::Widget> DoorSelDialog::DoorDDL(byte dir)
 	return DropDownList(
 		data = list_doortypes,
 		selectedValue = doors[dir],
-		fitParent = true,
-		width = 6_em,
```
EmilyV99: Fixed width here cut off some of the options; better to let it breathe. By removing `fitParent` I was able to make it still look the same visual style you had.

src/dialog/door_sel.cpp
```diff
@@ -46,34 +80,27 @@ std::shared_ptr<GUI::Widget> DoorSelDialog::view()
...
-	m = Map.CurrScr();
-	memcpy(doors, m->door, 4);
-	dcs = m->door_combo_set;
-	old_dcs = dcs;
-
 	window = Window(
 		title = "Select Door",
 		onClose = message::CANCEL,
 		Column(
-			Label(text = "Note: This only apply to 'NES Dungeon style' screens!", vAlign = 0, topPadding = 0.25_em, bottomPadding = 0.5_em),
```
EmilyV99: The default padding is usually good enough for most stuff

src/dialog/door_sel.cpp
```diff
@@ -92,33 +119,24 @@ bool DoorSelDialog::handleMessage(const GUI::DialogMessage<message>& msg)
 	{
 		case message::CLEAR:
 		{
-			for (int q = 0; q < 4; q++) doors[q] = dNONE;
+			for (int q = 0; q < 4; q++)
+				doors[q] = dNONE;
+			set_doors();
 			rerun_dlg = true;
-			return false;
-			break;
```
EmilyV99: A `break;` after a `return;` is useless, it'll never do anything.

src/dialog/door_sel.cpp
```diff
@@ -92,33 +119,24 @@ bool DoorSelDialog::handleMessage(const GUI::DialogMessage<message>& msg)
...
-				}
-			}
-			if (edited)
-			{
-				for (int q = 0; q < 4; q++)
-					m->door[q] = doors[q];
-				m->door_combo_set = dcs;
-				saved = false;
-			}
+			Map.FinishListCommand();
```
EmilyV99, line 130: "Finishing" the list command "confirms" the changes that were added to it, and now `Undo` will undo them.

src/dialog/door_sel.cpp
```diff
@@ -92,33 +119,24 @@ bool DoorSelDialog::handleMessage(const GUI::DialogMessage<message>& msg)
...
+			saved = false;
+			return true;
 		}
-		[[fallthrough]];
 		case message::CANCEL:
 		default:
 		{
+			Map.RevokeListCommand();
```
EmilyV99, line 137: ...and by revoking the list command, we instead cancel it, reverting any of it's changes.

src/dialog/door_sel.h
```diff
@@ -14,24 +14,26 @@ class DoorSelDialog : public GUI::Dialog<DoorSelDialog>
 {
 public:
 	enum class message {
-		OK, CANCEL, CLEAR
+		REFR_INFO, OK, CANCEL, CLEAR
```
EmilyV99, line 17: All dialogs should have `REFR_INFO`, the `INFOBTN` macro uses it.

src/zq/zq_doors.cpp
```diff
@@ -30,7 +30,7 @@ word door_combo_set_count;
 int32_t onDoors()
 {
     call_doorseldialog();
-    refresh(rMAP + rCOMBOS);
+    refresh(rMAP | rCOMBOS);
```
EmilyV99, line 33: `|` is better than `+` when dealing with bitflags

EmilyV99 (approved):
Fixed up the issues, and typed up comments here explaining everything I did so you can read over it @Jambu2024

PR #1099, opened 2025-06-07
fix(zscript): suspending ffc scripts no longer suspends screen scripts

Jambu2024:


connorjclark:
This does need a compat QR - I ran against the uploaded replays and found a quest that this broke.

Jambu2024:
Verywell give me a second to patch that in connor.

Jambu2024:
QR is done and All checks passing. should be ready to go.

connorjclark (approved):

PR #1103, opened 2025-06-09
feat: increase max items, sprites, sfx, midis, and dropsets from 256 to 512

Jambu2024:


src/items.h
```diff
@@ -276,6 +276,101 @@ struct itemdata
 		return;
 	}
 	void advpaste(itemdata const& other, bitstring const& pasteflags);
+	void clear(int16_t id)
+	{
+        tile = 0;
```
connorjclark: you should be able to instead do `*this = {};`

src/base/qst.cpp
```diff
@@ -9391,7 +9391,11 @@ int32_t readitems(PACKFILE *f, word version, word build)
 					break;
 			}
 		}
-		
+		if (s_version < 61) //Increased Number of Maximum Items
+		{
+			tempitem.clear(i);
```
connorjclark: why is this needed? the top of the loop does `tempitem = itemdata();`

Jambu2024: the function itself: cause im adding a new button to the Item Editor that uses it soon.
why is it needed here: its not. unless the constructor doesnt set the name appropriately.

Jambu2024:
still failing to load init data. not sure how to fix it. I tried somestuff but it didn't work. shrug @connorjclark you think you can finish this up

src/base/zdefs.h
```diff
@@ -146,17 +146,17 @@ enum {ENC_METHOD_192B104=0, ENC_METHOD_192B105, ENC_METHOD_192B185, ENC_METHOD_2
...
 #define V_DOORS            1
-#define V_ITEMS           60
-#define V_WEAPONS          8
+#define V_ITEMS           61
+#define V_WEAPONS          9
 #define V_COLORS           4 //Misc Colours
 #define V_ICONS            10 //Game Icons
 #define V_GRAPHICSPACK     1
-#define V_INITDATA        42
+#define V_INITDATA        63
```
EmilyV99: uhh, that's a big bump?

src/zq/zq_class.cpp
```diff
@@ -7625,8 +7625,8 @@ int32_t writedmaps(PACKFILE *f, word version, word build, word start_dmap, word
                 new_return(27);
             }
             
-            byte disabled[32];
-            memset(disabled,0,32);
+            byte disabled[64];
+            memset(disabled,0,64);
```
EmilyV99:
```suggestion
            byte disabled[MAXITEMS/8];
            memset(disabled,0,sizeof(disabled));
```

src/zc/ffscript.cpp
```diff
@@ -3928,7 +3928,7 @@ int32_t get_register(int32_t arg)
 				ret = -10000;
 				break;
 			}
-			ret=(itemsbuf[ri->idata].usedefence)*10000;
+			ret=(itemsbuf[ri->idata].usedefense)*10000;
```
EmilyV99:
```suggestion
			ret=(itemsbuf[ri->idata].usedefence)*10000;
```

src/zc/ffscript.cpp
```diff
@@ -15392,7 +15392,7 @@ void set_register(int32_t arg, int32_t value)
 				scripting_log_error_with_context("Invalid itemdata access: {}", ri->idata);
 				break;
 			}
-			(itemsbuf[ri->idata].usedefence)=vbound(value/10000, 0, 255);
+			(itemsbuf[ri->idata].usedefense)=vbound(value/10000, 0, 255);
```
EmilyV99:
```suggestion
			(itemsbuf[ri->idata].usedefence)=vbound(value/10000, 0, 255);
```

src/zc/ffscript.cpp
```diff
@@ -42123,7 +42123,7 @@ void FFScript::write_items(PACKFILE *f, int32_t vers_id)
 			{
 				Z_scripterrlog("do_savegamestructs FAILED to read ITEM NODE: %d",49);
 			}
-		if(!p_putc(itemsbuf[i].usedefence,f))
+		if(!p_putc(itemsbuf[i].usedefense,f))
```
EmilyV99:
```suggestion
		if(!p_putc(itemsbuf[i].usedefence,f))
```

src/zc/ffscript.cpp
```diff
@@ -42538,7 +42538,7 @@ void FFScript::read_items(PACKFILE *f, int32_t vers_id)
 			{
 				Z_scripterrlog("do_savegamestructs FAILED to write ITEM NODE: %d",49);
 			}
-		if(!p_getc(&itemsbuf[i].usedefence,f))
+		if(!p_getc(&itemsbuf[i].usedefense,f))
```
EmilyV99:
```suggestion
		if(!p_getc(&itemsbuf[i].usedefence,f))
```

src/zq/zq_class.cpp
```diff
@@ -8564,7 +8564,7 @@ int32_t writeitems(PACKFILE *f, zquestheader *Header)
 			{
 				new_return(49);
 			}
-			if(!p_putc(itemsbuf[i].usedefence,f))
+			if(!p_putc(itemsbuf[i].usedefense,f))
```
EmilyV99:
```suggestion
			if(!p_putc(itemsbuf[i].usedefence,f))
```

EmilyV99 (changes requested):
these are busted since they rely on the other PR

src/base/general.h
```diff
@@ -144,11 +144,11 @@ int wrap(int x,int low,int high);
 //The offset from dmap/mapscr-> midi/screen_midi to currmidi
 #define MIDIOFFSET_DMAP            (ZC_MIDI_COUNT-4)
 #define MIDIOFFSET_MAPSCR          (ZC_MIDI_COUNT-4)
-//The offset from currmidi to ZScript MIDI values
+//The offset from currmidi to0 ZScript MIDI values
```
EmilyV99:
```suggestion
//The offset from currmidi to ZScript MIDI values
```

src/zq/zq_class.cpp
```diff
@@ -7527,7 +7527,7 @@ int32_t writedmaps(PACKFILE *f, word version, word build, word start_dmap, word
                 new_return(10);
             }
             
-            if(!p_putc(DMaps[i].midi,f))
+            if(!p_iputl(DMaps[i].midi,f))
```
EmilyV99:
```suggestion
            if(!p_iputw(DMaps[i].midi,f))
```
int16_t is `iputw` not `iputl`

src/base/mapscr.h
```diff
@@ -144,12 +144,12 @@ struct mapscr
...
-	byte bosssfx;
-	byte secretsfx = 27;
-	byte holdupsfx = 20;
+	int16_t oceansfx;
+	int16_t bosssfx;
+	int16_t secretsfx = 27;
+	int16_t holdupsfx = 20;
 
-	int16_t screen_midi = -1;
+	int32_t screen_midi = -1;
```
EmilyV99:
```suggestion
	int16_t screen_midi = -1;
```
...why int32_t? we don't need to go above 16 bits here for sure, there won't be >32k midis in a single quest...

src/zq/zq_class.cpp
```diff
@@ -9316,17 +9316,17 @@ int32_t writemapscreen(PACKFILE *f, int32_t i, int32_t j)
...
-		if(!p_putc(screen.secretsfx,f))
+		if(!p_iputw(screen.secretsfx,f))
 			return qe_invalid;
-		if(!p_putc(screen.holdupsfx,f))
+		if(!p_iputw(screen.holdupsfx,f))
 			return qe_invalid;
 		if(!p_iputw(screen.timedwarptics,f))
 			return qe_invalid;
-		if(!p_iputw(screen.screen_midi,f))
+		if(!p_iputl(screen.screen_midi,f))
```
EmilyV99:
```suggestion
		if(!p_iputw(screen.screen_midi,f))
```
See other comment, this shouldn't need to be changed? screen_midi should just be int16_t

src/zq/zq_class.cpp
```diff
@@ -11505,10 +11505,10 @@ int32_t writeguys(PACKFILE *f, zquestheader *Header)
 			{
 				if (!p_iputw(guysbuf[i].burnsprs[q], f))
 					new_return(116);
-				if (!p_iputw(guysbuf[i].light_rads[q], f))
+				if (!p_putc(guysbuf[i].light_rads[q], f))
```
EmilyV99, line 11508: ...Oh, this was straight up just wrong before. Nice catch!

src/zq/zq_class.cpp
```diff
@@ -14170,7 +14170,7 @@ int32_t writeinitdata(PACKFILE *f, zquestheader *)
 		writesize=0;
 		
 		for(int q = 0; q < MAXITEMS/8; ++q)
-			if(!p_putc(zinit.items[q], f))
+			if(!p_iputw(zinit.items[q], f))
```
EmilyV99:
```suggestion
			if(!p_putc(zinit.items[q], f))
```
Wrong change- this is an array of bytes, it will always be `putc`- the size of the array is what changes, not the size of each index.

src/base/qst.cpp
```diff
@@ -4785,14 +4797,27 @@ int32_t readdmaps(PACKFILE *f, zquestheader *Header, word, word, word start_dmap
 		
 		if(s_version>2)
 		{
-			byte di[32];
-			
-			if(!pfread(&di, 32, f)) return qe_invalid;
+			byte di[MAXITEMS/8];
 			
-			for(int32_t j=0; j<MAXITEMS; j++)
+			if (s_version >= 23)
```
EmilyV99, line 4802: instead of doing an if/else here, you should just do like, an `items_to_read = s_version >= 23 ? MAXITEMS : 256;`, to duplicate less code

src/base/qst.cpp
```diff
@@ -6501,58 +6554,116 @@ int32_t readitems(PACKFILE *f, word version, word build)
...
-                        {
-                            return qe_invalid;
-                        }
-                    }
+					if (s_version >= 62)
+					{
+						if (!p_igetw(&tempitem.wpn, f))
+						{
+							return qe_invalid;
+						}
```
EmilyV99, line 6562:
```suggestion
							return qe_invalid;
```
Should probably not brace these to help reduce line count (there's a lot of these.... it adds up)

src/base/qst.cpp
```diff
@@ -6691,17 +6802,30 @@ int32_t readitems(PACKFILE *f, word version, word build)
...
                     }
                     
-                    if(!p_getc(&tempitem.usesound,f))
-                    {
-                        return qe_invalid;
-                    }
+					if (s_version >= 62)
+					{
+						if (!p_igetw(&tempitem.usesound, f))
+							return qe_invalid;
```
EmilyV99, line 6808: you should probably put the `usesound2` also in this `if`, and then  handle the old usesound2 stuff in the `else`

src/base/qst.cpp
```diff
@@ -7028,8 +7162,8 @@ int32_t readitems(PACKFILE *f, word version, word build)
 
 	if (should_skip)
 		return 0;
-    
-    //////////////////////////////////////////////////////
+
+	//////////////////////////////////////////////////////
     // Now do any updates because of new item additions
```
EmilyV99, line 7167: If you're gonna retab the first part of this comment, you should do the rest too- now it's misaligned

src/base/qst.cpp
Jambu2024: not for this... as it needs to be an constant value. I tried the following didn't work.
image

src/base/qst.cpp
Jambu2024: image
looks fine on my end. not sure how to even fix that

src/base/qst.cpp
Jambu2024: done

src/base/qst.cpp
EmilyV99: It only looks fine because your editor has tabs set to display as the same number of spaces that happen to be there.
Just change the spaces on all those lines to tabs at the start instead. It's mis-indented.

src/base/qst.cpp
EmilyV99: yeah.... could use `malloc` to create a non-static sized buffer, but it might just not be worth the extra complication. OK to just leave it then, my bad.

src/base/qst.cpp
Jambu2024: done

Jambu2024:
yeah i cannot figure out how to expand midi to 512. need some help with that

Alucard648:
Is it ever possible to do this? A guy crashed hard into SFX limit on attempt to add voice acting to his quest.

EmilyV99:
> Is it ever possible to do this? A guy crashed hard into SFX limit on attempt to add voice acting to his quest.

That sounds like something that would likely be problematic *even if we DID expand this* - the more size you put in SFX, the longer it takes to load them- and they need to load every single time you load the quest. I know SLG had a ton of LARGE sfx in her quest at one point, and it literally took multiple minutes after I pressed start on the title screen before I got to the opening wipe, all just loading SFX (given, that might've been on a slower build / PC, but still, it's an example of the issue).

What we'd need to add for a quest that wants PROPER large banks of sfx would be the ability to have scripts play an SFX from a file the same way enhanced music works now, which is a whole other thing. But that'd be very hard to do for things outside of scripts for sure, so...

PR #1105, opened 2025-06-11
fix(zc): prevent softlock when enemy falls into pit after eating hero

EmilyV99:
…the hero


connorjclark (approved):
i guess we should fix this... would be kinda funny not to though.

PR #1110, opened 2025-07-01
Item editor moveflags

EmilyV99:


src/zc/scripting/types/npcdata.cpp
```diff
@@ -47,6 +47,20 @@ static ArrayRegistrar NPCDATABEHAVIOUR_registrar(NPCDATABEHAVIOUR, []{
 	return &impl;
 }());
 
+static ArrayRegistrar NPCDATAMOVEFLAGS_registrar(NPCDATAMOVEFLAGS, []{
+	static ScriptingArray_ObjectMemberBitwiseFlags<guydata, &guydata::moveflags, 16> impl;
+	impl.setDefaultValue(0);
+	impl.setMul10000(false);
```
connorjclark, line 53: all of these new arrays should be *10000 (`impl.setMul10000(true);`

PR #1111, opened 2025-07-03
feat: A bunch of new combo trigger stuff

EmilyV99:
- longer timer triggers
- triggers can be conditional on LevelStates and GlobalStates in either direction
- triggers can un-set an ExState or ExDoor
- triggers can use "special counters" (like the subscreen, most importantly the "Any Keys")
- Triggers can play strings. Also they can do a different string (and different button prompt) if their "conditions" currently fail.
- The "screen load" trigger is now recursive- so if a "screen load" trigger changes the combo to another combo with a "screen load" trigger that will trigger too.
- The ability to make the player jump. (Or well, to set the player's Jump value, so it could work for like, a fan slowly blowing you into the air, or to shove you down towards the ground with negatives)
- The conditional to only trigger if the player is "standing", or "not standing" (works for sideview as well)
- The conditional to only trigger if the player's Z value is `>=` or `<` a set value.
- GUI re-organization and cleanup for the trigger gui


resources/include/bindings/combotrigger.zh
```diff
@@ -156,16 +161,16 @@ class combotrigger {
...
-	// Used as a minimum (or maximum, if [TRIGFLAG_INVERT_MIN_MAX] is set) level
+	// Used as a minimum (or maximum, if [ComboTriggerFlag->TRIGFLAG_INVERT_MIN_MAX|TRIGFLAG_INVERT_MIN_MAX] is set) level
 	// for any weapon-based triggers.
 	//
 	// @zasm_var CMBTRIGWPNLEVEL
 	internal int WeaponLevel;
 	
 	// If set to non-0, represents an item that must be owned (or must NOT
-	//   be owned, if [TRIGFLAG_INVERT_ITEMREQ] is set).
+	//   be owned, if [ComboTriggerFlag->TRIGFLAG_INVERT_ITEMREQ|TRIGFLAG_INVERT_ITEMREQ] is set).
```
connorjclark: Are you making this change with the docs or the extension in mind? What is it solving?

If we want to display enums in a particular way, it should be done in the comment formatter (`parseForSymbolLinks`). Otherwise this will be inconsistently applied. It's also just kind of noisy / boilerplate-y.

Also, this suggests that `ComboTriggerFlag->TRIGFLAG_SWORD` is valid code but it isn't (enums aren't scoped in the language, just globals), I'd suggest something more like (`TRIGFLAG_SWORD (enum ComboTriggerFlag)`) (but again, not in the source code text, but just as how enum values get formatted in the comment parser)

resources/include/bindings/combotrigger.zh
```diff
@@ -208,25 +213,25 @@ class combotrigger {
...
+	// With [ComboTriggerFlag->TRIGFLAG_COUNTER_EAT|TRIGFLAG_COUNTER_EAT], consumes the [combotrigger->CounterAmount|CounterAmount] from the counter when triggered.
+	// ...and with [ComboTriggerFlag->TRIGFLAG_COUNTER_NOT_ONLY_TRIG_EAT|TRIGFLAG_COUNTER_NOT_ONLY_TRIG_EAT], the consumed amount consumes even if the combo
 	// fails to trigger (ex. via failing a conditional)
 	//
 	// @zasm_var CMBTRIGCTR
 	internal int Counter;
 	
-	// The counter amount used by this trigger. See [Counter].
+	// The counter amount used by this trigger. See [combotrigger->Counter|Counter].
 	// @zasm_var CMBTRIGCTRAMNT
```
connorjclark, line 235: same as above. We can change the comment formatter to print the field w/ the context of the parent object type, we don't have to label everything like that manually.

src/base/combo.h
```diff
@@ -317,9 +321,11 @@ struct newcombo
...
 //triggerflags[4]
 #define combotriggerSCREENLOAD         0x00000001
-// #define combotrigger                    0x00000002
-// #define combotrigger                    0x00000004
-// #define combotrigger                    0x00000008
+#define combotriggerUNSETEXSTATE       0x00000002
+#define combotriggerUNSETEXDOOR        0x00000004
+#define combotriggerPLAYER_STANDING    0x00000008
+#define combotriggerPLAYER_NOTSTANDING 0x00000010
+#define combotriggerINVERT_PLAYER_Z    0x00000020
```
connorjclark, line 328: needs zscript entries in ComboTriggerFlag

resources/include/bindings/combotrigger.zh
connorjclark: (nvm, this is the symbol linking being buggy, not you changing how the text displays)

src/base/combo.h
EmilyV99: they indeed already have them, added in a later commit

PR #1112, opened 2025-07-06
Conveyor stuff

EmilyV99:
fixes need either a qr or replay check, and a backport. Unsure how to handle that.


connorjclark (approved):
uploaded replays all good still

PR #1114, opened 2025-07-09
fix(zc): bombs that are auto-lifted being immediately thrown

EmilyV99:
replay verison check / qr + backport?


connorjclark:
don't need to add any replay check for compat, all our replays are good–even the ~100 uploaded replays.

PR #1116, opened 2025-07-09
fix(zc): prevent scrolling over liquid/pits that should drown/fall you

EmilyV99:
replay verison check / qr + backport?


connorjclark:
let's do a qr for this one

PR #1121, opened 2025-07-10
arrays / strings + GC

connorjclark:
See the commit descriptions.


src/parser/CompileError.xtable
```diff
@@ -125,7 +125,7 @@ X(	BadAnnotation,			P,	A,	E,	STR,	STR,	"Annotation @%s invalid: %s"						)
 X(	BadAutoType,			S,	A,	E,	STR,	STR,	"'%s' %s"							)
 X(	GroupAuto,			S,	A,	E,	VOID,	VOID,	"'auto' cannot apply to a group declaration."							)
 X(	BadReturnType,			S,	A,	E,	STR,	VOID,	"Cannot return type '%s'."							)
-X(	BadTempVal,			T,	A,	E,	VOID,	VOID,	"Temporary literals cannot be stored!"								)
+X(	DEFUNCT_BadTempVal,			T,	A,	E,	VOID,	VOID,	"Temporary literals cannot be stored!"								)
```
EmilyV99:
```suggestion
X(	BadTempVal,			T,	D,	E,	VOID,	VOID,	"Temporary literals cannot be stored!"								)
```
there's already a way to deprecate stuff built-in here it looks like

resources/headers/ghost_zh/2.8/ghost2_other.zh
EmilyV99, line 1: Nice job here, the class is a much cleanear solution for this!

src/parser/BuildVisitors.h
```diff
@@ -123,27 +121,14 @@ namespace ZScript
...
 		std::vector<std::shared_ptr<Opcode>> result;
 		int32_t returnlabelid;
-		int32_t returnRefCount;
 		std::vector<int32_t> continuelabelids;
-		std::vector<int32_t> continueRefCounts;
 		std::vector<int32_t> breaklabelids;
-		std::vector<int32_t> breakRefCounts;
 		std::vector<std::vector<Scope*>> breakScopes;
-		std::vector<uint> break_past_counts;
 		std::vector<uint> break_to_counts;
```
EmilyV99, line 131: `continue_to_counts` removed but `break_to_counts` still here? is that a mistake?
```suggestion
```

docs-www/source/zscript/lang/declarations/arrays_strings.rst
```diff
@@ -179,8 +188,3 @@ which does store it for longer-term use.
 	printf(str); // outputs 'Test String'
 
 	char32[] strs[] = { <error>"Test 1"</error>, <error>"Test 2"</error> }; // Error T100: Temporary literals cannot be stored
-
-.. note::
-	:ref:`'#include' statements<directive_import>` use a special form of
-	string literal with slightly different rules (notably, not using escape
-	characters)
```
EmilyV99: Did this change? I know some users would be unhappy with this (people like their 'C:\\windows\style\paths' apparently)

tests/scripts/errors/errors_7.zs
```diff
@@ -38,9 +38,7 @@ void test()
 	untyped foo = arr;
 	foo = arr2;
 	foo = enemies;
-
-	//!TODO
-	arr2[1] = {2,3,4}; //!TODO should be valid, if arrays become gc-managed
+	arr2[1] = {2,3,4};
```
EmilyV99, line 41: If this isn't meant to error I'd remove it from this file entirely and put it in a different test
```suggestion
```

src/zc/scripting/types/game.cpp
```diff
@@ -1299,3 +1299,19 @@ static ArrayRegistrar GAMEGRAVITY_registrar(GAMEGRAVITY, []{
...
+
+static ArrayRegistrar GAME_SAVED_PORTALS_registrar(GAME_SAVED_PORTALS, []{
+	static ScriptingArray_GlobalComputed<int> impl(
+		[](int){ return game->user_portals.size(); },
+		[](int, int index) -> int {
+			return game->user_portals[index].getUID();
+		},
+		[](int, int index, int value){
+			return false;
+		}
```
EmilyV99, line 1311: Could we have a function declared somewhere that is just this, that can be shared by read-only arrays like this?

tests/scripts/playground/auto/garbage_collection.zs
```diff
@@ -59,6 +59,24 @@ class Empty
...
 }
 
+class SelfRetainingRng
+{
+	SelfRetainingRng self;
+	randgen rng;
+
+	SelfRetainingRng(randgen rng)
+	{
+		this->self = self;
```
EmilyV99:
```suggestion
		this->self = this;
```
I'm uh, pretty sure this was meant to be `this`?

tests/scripts/playground/auto/garbage_collection.zs
```diff
@@ -715,63 +689,168 @@ generic script garbage_collection
...
 			// GC runs when needed to make room.
 			for (int i = 0; i < 300; i++)
 			{
 				randgen rng = Game->LoadRNG();
-				if (<untyped>rng == 0)
-					printf("failed to allocate randgen\n");
+				if (rng == NULL)
+					Test::Fail("failed to allocate randgen");
+
+				auto retainer = new SelfRetainingRng(rng); // Only gets deleted after a full GC.
```
EmilyV99, line 710: This test didn't catch the error I pointed out, so could use some expansion to cover that case.

tests/scripts/playground/auto/garbage_collection.zs
```diff
@@ -924,4 +1002,24 @@ generic script garbage_collection
...
+	void storeVarargs(...int[] varargs)
+	{
+		check("RefCount(varargs)", RefCount(varargs), 1L);
+		storedVarargs = varargs;
+	}
+
+	char32[] makeString(int len)
+	{
+		char32[] s = "";
+		for (int i = 0; i < len; i++) ArrayPushBack(s, '.');
```
EmilyV99:
```suggestion
		for (int i = 0; i < len; i++) ArrayPushFront(s, '.');
```
Pushing back here will add the characters after the `\0`, which, is not properly appending to the "string". Since all chars are identical, pushing the front works instead- alternately using the string functions such as `strcat` would append properly, but that's overkill here.

tests/scripts/playground/auto/garbage_collection.zs
```diff
@@ -715,63 +689,168 @@ generic script garbage_collection
...
+			check("RefCount(items)", RefCount(items), 1L);
+
+			checkCountWithGC(1);
+		}
+		checkCountWithGC(0);
+
+		printf("=== Test %d - return local string === \n", ++tests);
+		{
+			auto s = makeString(5);
+			check("RefCount(s)", RefCount(s), 1L);
```
EmilyV99, line 784:
```suggestion
			check("RefCount(s)", RefCount(s), 1L);
			Test::AssertEqual(strlen(s), 5); // strlen accounts for \0 termination
```

tests/scripts/playground/auto/garbage_collection.zs
```diff
@@ -715,63 +689,168 @@ generic script garbage_collection
...
+		}
+
+		printf("=== Test %d - string literal === \n", ++tests);
+		{
+			char32[] str = "hi there\n";
+			check("RefCount(str)", RefCount(str), 1L);
+			Trace(str);
+
+			check("RefCount(\"literal\")", RefCount("literal"), 1L);
+		}
```
EmilyV99, line 802:
```suggestion
			check("RefCount(\"literal\")", RefCount("literal"), 1L);
			checkCountWithGC(1); // literal doesn't stay allocated, since not stored
		}
		checkCountWithGC(0);
```

tests/scripts/playground/auto/arrays.zs
```diff
@@ -70,5 +88,62 @@ generic script arrays
...
+		auto ComboD = Screen->ComboD;
+		Screen->ComboD[0] = 100;
+		ComboD[0]++;
+		Test::AssertEqual(ComboD[0], 101);
+		// TODO: negative indices currently do not work for internal arrays.
+		// Need to:
+		// 1. implement in each IScriptingArray's getElement/setElement
+		// 2. figure out what to do about "index bounding" internal arrays.
+		//    Maybe add a QR for neg. index internal arrays, and have that
+		//    avoid the existing index bounding behavior.
```
EmilyV99, line 116: Hmm, yeah makes sense that'd be trouble.

EmilyV99 (changes requested):

src/parser/BuildVisitors.h
connorjclark: `BuildOpcodes::caseStmtWhile` still uses it, though trivially, but I didn't take the time to rip it out. Left a note that it could be done later.

docs-www/source/zscript/lang/declarations/arrays_strings.rst
connorjclark: It did not change, I only removed it b/c while correct, it didn't seem too useful to document. But I'll just put it back.

tests/scripts/playground/auto/garbage_collection.zs
connorjclark: yup, thx

tests/scripts/playground/auto/garbage_collection.zs
connorjclark: these helpers track `count` which only change w/ the Person class

tests/scripts/errors/errors_7.zs
connorjclark: there's lots of snippets in these "error" files that are also checking the negative case (shouldn't error), so i'll keep it here

PR #1122, opened 2025-07-17
feat: qr for "Weapons Cannot Stunlock Enemies (ignore)"

EmilyV99:
Like the existing no stunlock rule, stops stunned enemies from being hit. Unlike the existing rule, enemies IGNORE instead of BLOCK the weapon. The existing qr was renamed adding "(block)" to the end to differentiate.


Alucard648:
AfaiK, there are 2 odd enemy defense types called Stun or Ignore and Stun or Block. They are supposed to allow defining stunlock vulnerability and handling weapon interaction on attempt to stunlock. But something makes them borked....

EmilyV99:
> AfaiK, there are 2 odd enemy defense types called Stun or Ignore and Stun or Block. They are supposed to allow defining stunlock vulnerability and handling weapon interaction on attempt to stunlock. But something makes them borked....

That is not what those do at all. They have nothing to do with stunlocking. "Stun or Ignore" is "Power 0 weapons are ignored, everything else stuns". "Stun or Block" is "Power 0 weapons are blocked, everything else stuns".

PR #1124, opened 2025-07-19
chore!: add cstdint include in src/parser/CommentUtils.h

thican:
Fails to compile with GCC without the `<cstdint>` include.


connorjclark:
thanks!

PR #1132, opened 2025-08-07
Enable building without GTK

jggatter:
Hi,

I reached out earlier this week to see if it was possible to build the `zplayer` without GTK dependencies. I didn't give much context at the time, but my interest is in seeing if I can get the player to work on handheld Linux ARM devices and one of the runtimes supported by the [PortMaster](https://portmaster.games/) project. Coincidentally it looks like that community had [looked into the feasibility](https://suggestions.portmaster.games/suggestion-details?id=b06f2329b04040c995f24b06a01bf4f4) of porting ZQuest Classic before. I'm new to this, but it looks like the [Westonpack runtime](https://github.com/binarycounter/Westonpack/wiki) + Box64 is what I would use, and the PostMaster community generally describes GTK as incompatible. So my goal here was to see if I could build the project without GTK-dependent stuff like NFD, which seemed to be possible given the cool work done here in creating a browser runtime.

Massive thanks for disabling NFD inclusion in the `CMakeLists.txt `and `src/files.cpp`! This was tremendously helpful in showing me how I might do this for the rest of the project. I continued to make progress here by doing the same in `src/base/zsys.cpp` and the `zconsole`. I also built upon a previous issue that I saw referenced in the `CMakeLists.txt` (https://github.com/liballeg/allegro5/issues/1335) that would build Allegro without need of GTK3: Disabling NFD and (not sure if necessary but) enabling the SDL2 experimental backend.

Ultimately with a lot of troubleshooting I was able to produce something that builds and runs within an amd64 Docker container.
<details>

<summary>Dockerfile and run script</summary>

```dockerfile
FROM ubuntu:24.04

# Update package list and install dependencies
RUN apt update -y \
    && apt install -y \
        # Build tools
        build-essential \
        gcc-multilib \
        g++-multilib \
        cmake \
        clang \
        ninja-build \
        git \
        flex using\
        bison \
        pkg-config \
        zlib1g-dev \
        uuid-dev \
        elfutils \
        libcurl4-openssl-dev \
        # Graphical libraries
        libsdl2-dev \
        libsdl2-mixer-dev \
        libpng-dev \
        libjpeg-dev \
        libx11-dev \
        libglu1-mesa-dev \
        freeglut3-dev \
        mesa-common-dev \
        libopengl0 \
        libglu1 \
        libxcursor1 \
        # Audio libraries
        libasound2-dev \
        libdumb1-dev \
        libogg-dev \
        libvorbis-dev

# Copy project files into the container
COPY . /opt/zquestclassic

# Set working directory
WORKDIR /opt/zquestclassic

# Configure and build the project
RUN CC=clang-18 CXX=clang++-18 cmake \
    -G 'Ninja Multi-Config' \
    -B build \
    -DWANT_NFD=FALSE \
    -S .

RUN cmake --build build --config Debug -t zplayer -j 8

WORKDIR /opt/zquestclassic/build/Debug
CMD ["./zplayer"]
```
```sh
#!/bin/bash
# Run container on Fedora 42 KDE

set -e

docker run -it --rm \
  -e DISPLAY=$DISPLAY \
  -e PULSE_SERVER=unix:${XDG_RUNTIME_DIR}/pulse/native \
  -v /tmp/.X11-unix:/tmp/.X11-unix \
  -v ${XDG_RUNTIME_DIR}/pulse/native:${XDG_RUNTIME_DIR}/pulse/native \
  -v $HOME/.config/pulse/cookie:/root/.config/pulse/cookie \
  -v $PWD/quests:/opt/zquestclassic/build/Debug/quests \
  --device /dev/dri \
  --group-add video \
  zquestclassic
```
</details>

This is my first project working with graphics/audio dependencies, so I'm still working on my understanding of these things. Please let me know whether I'm going about this the wrong way!

I'll see if I can get the build outputs to work on my ARM handheld.

Thanks,
James

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,6 +317,11 @@ if(NOT allegro5_POPULATED)
 	# We don't support FLAC.
 	option(WANT_FLAC "" OFF)
 	option(WANT_VIDEO "" OFF)
+
+	if(NOT USE_NFD)
+	  set(WANT_NATIVE_DIALOG OFF CACHE BOOL "" FORCE)
+	  set(ALLEGRO_SDL ON CACHE BOOL "" FORCE)
+	endif()
   add_subdirectory(${allegro5_SOURCE_DIR} ${allegro5_BINARY_DIR} EXCLUDE_FROM_ALL)
   # Allegro 5 CMakeLists sets CMAKE_CONFIGURATION_TYPES in a way that doesn't work for us,
   # so set it back to our value.
@@ -365,7 +370,7 @@ target_include_directories(allegro_with_legacy INTERFACE ${allegro5_SOURCE_DIR}/
 target_include_directories(allegro_with_legacy INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/allegro_legacy/include ${CMAKE_CURRENT_BINARY_DIR}/third_party/allegro_legacy/include)
 target_link_libraries(allegro_with_legacy INTERFACE allegro-legacy allegro_primitives allegro_font allegro_image allegro_audio allegro_acodec allegro_main)
 
-if (NOT EMSCRIPTEN)
+if (USE_NFD AND NOT EMSCRIPTEN)
 	target_link_libraries(allegro_with_legacy INTERFACE allegro_dialog)
 endif()
 
@@ -643,6 +648,9 @@ endif()
 if (USE_NFD)
 	target_link_libraries(zcbase PUBLIC nfd)
 	target_compile_definitions(zcbase PUBLIC HAS_NFD)
+else()
+	find_package(SDL2 REQUIRED)
+	target_link_libraries(zcbase PUBLIC SDL2::SDL2)
 endif()
 
 # std::execution supplementation.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -751,6 +751,10 @@ add_library(zconsolelogger STATIC src/zconsole/ConsoleLogger.cpp)
 target_link_libraries(zconsolelogger PUBLIC allegro_with_legacy zcbase)
 target_include_directories(zconsolelogger PUBLIC src)
 
+if(USE_NFD)
+	target_compile_definitions(zconsolelogger PUBLIC HAS_NFD)
+endif()
+
 target_link_libraries(zcbase PUBLIC zconsolelogger)
 
 #############################################################
--- a/src/base/zsys.cpp
+++ b/src/base/zsys.cpp
@@ -226,7 +226,7 @@ char datapwd[8]   = "longtan";
 #if defined(ALLEGRO_MAXOSX)
     printf("%s",buf);
 #endif
-#ifndef __EMSCRIPTEN__
+#ifdef HAS_NFD
     if (!zscript_coloured_console.valid() && !is_headless())
     {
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -317,6 +317,11 @@ if(NOT allegro5_POPULATED)
 	# We don't support FLAC.
 	option(WANT_FLAC "" OFF)
 	option(WANT_VIDEO "" OFF)
+
+	if(NOT USE_NFD)
+	  set(WANT_NATIVE_DIALOG OFF CACHE BOOL "" FORCE)
+	  set(ALLEGRO_SDL ON CACHE BOOL "" FORCE)
```
connorjclark, line 323: You should be able to configure your build w/ `-DALLEGRO_SDL=ON` and `-DWANT_NATIVE_DIALOG=OFF` without these changes – does that work for you?

CMakeLists.txt
```diff
@@ -643,6 +648,9 @@ endif()
 if (USE_NFD)
 	target_link_libraries(zcbase PUBLIC nfd)
 	target_compile_definitions(zcbase PUBLIC HAS_NFD)
+else()
```
connorjclark, line 651: This should be unnecessary if configuring w/ `-DALLEGRO_SDL=ON` - that's why the web build does (see `configure_emscripten.sh`, and allegro's cmake will then handle actually pulling in SDL2.

CMakeLists.txt
```diff
@@ -365,7 +370,7 @@ target_include_directories(allegro_with_legacy INTERFACE ${allegro5_SOURCE_DIR}/
 target_include_directories(allegro_with_legacy INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/allegro_legacy/include ${CMAKE_CURRENT_BINARY_DIR}/third_party/allegro_legacy/include)
 target_link_libraries(allegro_with_legacy INTERFACE allegro-legacy allegro_primitives allegro_font allegro_image allegro_audio allegro_acodec allegro_main)
 
-if (NOT EMSCRIPTEN)
+if (USE_NFD AND NOT EMSCRIPTEN)
```
connorjclark, line 373: nfd is unrelated to the allegro dialog addon (which is used in Linux and OSX for a basic text log UI), so I don't like using it here.

Should be OK to define a new option in our cmake for `WANT_NATIVE_DIALOG` (default to on, set `HAS_ALLEGRO_DIALOG` compile definition on zcbase). It's the same name as the option for Allegro, but in our cmake we just use it conditionally link against allegro_dialog.

connorjclark (commented):
Thanks for sharing those details, happy to answer any questions to help you figure out your porting effort.

I'd like to have more direct configuration options – it's a little confusing that disabling the nfd library (native file dialogs) would opt you into the SDL backend, or out of the allegro dialog addon (which we use for a basic text view on OSX and Linux). Additionally, though the motivation is to avoid depending on GTK – that's not a concern on our other platforms, yet the way these options are being applied don't take that into consideration.

Left a few specific comments.

connorjclark:
FYI, looks like GTK is used in allegro on linux except for some platforms. I don't think there's a way to control that in [their cmake](https://github.com/liballeg/allegro5/blob/efb3129bb54ecfbfbc59984645b948d86766f23d/CMakeLists.txt#L366), perhaps you could request they add one.

PR #1134, opened 2025-08-14
Cutscene effect combo

EmilyV99:
https://discord.com/channels/876899628556091432/1405051996750217267


src/dialog/combowizard.cpp
```diff
@@ -933,7 +942,22 @@ void combo_default(newcombo& ref, bool typeonly)
...
 		case cLOCKBLOCK2: case cBOSSLOCKBLOCK2:
 			break;
-		//
+		case cCUTSCENEEFFECT:
+		{
+			switch(ref.attribytes[0])
+			{
+				default:
+					ref.attribytes[0] = CUTEFF_PLAYER_WALK;
+				[[fallthrough]];
```
connorjclark, line 951: i'd just put this before the switch (but nbd)

src/zc/hero.cpp
```diff
@@ -284,31 +290,31 @@ int32_t getCurrentActiveShield()
 int32_t refreshActiveShield()
 {
 	int32_t id = -1;
-    if(DrunkcBbtn())
+    if(getHeroInput(btnB, false, true))
```
connorjclark: more than one boolean in parameter is a bad smell. tough to read this code as handling "drunk" code. A separate "getDrunkHeroInput" would _really_ help here.

connorjclark (approved):

src/dialog/combowizard.cpp
EmilyV99: it feels nicer using the `default:` for it so it blatantly checks for cases that are not present (instead of needing to manually check the value against CUTEFF_MAX above the switch)

src/zc/hero.cpp
EmilyV99: Handled via flags instead of bool params. (I personally think bool params would work fine... if they were keyword args like in python. But yeah, flags makes more sense for this here for sure)

PR #1136, opened 2025-08-18
try to fix pitfall sensitivity issues

EmilyV99:


EmilyV99:
#1162 attempts to supercede this

PR #1138, opened 2025-08-20
feat: minimap can draw visited rooms

EmilyV99:


src/base/zdefs.h
```diff
@@ -279,12 +279,12 @@ if(close_button_quit) \
...
-	warpfxWAVYIN, warpfxWAVYOUT, warpfxWAVYNOHERO, warpfxWAVYINNOHERO, warpfxWAVYOUTNOHERO, 
+//We have definitions for this below, but this is for scripted stuff in 2,54.
+enum { warpfxNONE, warpfxBLACKOUT, warpfxWIPE, warpfxSCROLL, warpfxZAP, warpfxZAPIN, warpfxZAPOUT, warpfxWAVY,
+	warpfxWAVYIN, warpfxWAVYOUT, warpfxWAVYNOHERO, warpfxWAVYINNOHERO, warpfxWAVYOUTNOHERO,
 		warpfxLAST};
-	
-//wipe types - mosaic should be one of them. 
+
+//wipe types - mosaic should be one of them.
 enum { wipeNONE, wipeCOOL, wipeBLACK, wipeBSZELDA, wipeTRIANGLE, wipeCIRCLE, wipeALLSTARS, wipeLAST };
```
EmilyV99, line 273: All of these random spacing changes shouldn't be here; I'm guessing your visual studio reformatted the whole file when you edited it or something @Deathrider365 ? Would be best to fix that before continuing and revert any of the whitespace changes (reverting specific sections of changes is relatively easy in the git gui)

src/new_subscr.cpp
```diff
@@ -3139,26 +3142,64 @@ void SW_MMap::draw(BITMAP* dest, int32_t xofs, int32_t yofs, SubscrPage& page) c
...
+			{
+				if (!(dl & 0x80))
+				{
+					dl <<= 1;
+					continue;
+				}
+
+				dl <<= 1;
+
+				int scrx = x2 - thedmap.xoff;
```
EmilyV99: per [this commit](https://github.com/ZQuestClassic/ZQuestClassic/commit/75c45c19ac01d6dfed0049e4369b17388622ef1c), this is a bug. Should be:
```suggestion
				int scrx = x2 + thedmap.xoff;
```

src/components/zasm/table.cpp
```diff
@@ -2766,6 +2766,7 @@ static constexpr script_variable variable_list[]=
 	{ "ENGINE_MUSIC_ACTIVE", ENGINE_MUSIC_ACTIVE, 0},
 	{ "CMBTRIG_PLAY_MUSIC", CMBTRIG_PLAY_MUSIC, 0},
 	{ "CMBTRIG_MUSIC_REFRESH", CMBTRIG_MUSIC_REFRESH, 0},
+	{ "SUBWIDGTY_COLOR_VISITED", SUBWIDGTY_COLOR_VISITED, 0 },
```
connorjclark, line 2769: missing change in `get_register_ref_dependency` (see all the other `SUBWIDGTY_COLOR` vars)

src/subscr.cpp
```diff
@@ -2279,58 +2279,59 @@ void frame2x2(BITMAP *dest,int32_t x,int32_t y,int32_t tile,int32_t cset,int32_t
...
-				if(dl&0x80)
-					rectfill(dest,x_1+x+2,y_1+y,x_2+x-2,y_2+y,c1);
-			}
-			
-			dl<<=1;
-		}
-		
-		++si;
-	}
+    if (unvis_color < 0 && bg_color < 0 && vis_color < 0)
```
connorjclark: Please do not reverse indentation - going the wrong direction here :) also just trickier to review the diff

src/subscr.cpp
```diff
@@ -2279,58 +2279,59 @@ void frame2x2(BITMAP *dest,int32_t x,int32_t y,int32_t tile,int32_t cset,int32_t
...
+            int scrx = x2 + thedmap.xoff;
+            if (scrx < 0x0 || scrx > 0xF)
+                continue;
+            int scrid = scrx + y2 * 0x10;
+
+            int x_1 = x2 * 8, x_2 = x_1 + 6;
+            int y_1 = y2 * 4, y_2 = y_1 + 2;
+            if (reg.get_region_id(scrid) != 0)
+            {
+                //bool top = scrid >= 0x10 && reg.is_same_region(scrid, scrid - 0x10);
```
connorjclark: why is top commented out?

src/new_subscr.cpp
```diff
@@ -3483,93 +3483,91 @@ byte SW_MMap::getType() const
...
-			{
-				rectfill(dest,tx+8,ty+8,tx+71,ty+39,c.dngn_bg);
+				int32_t maptile = (!get_qr(qr_BROKEN_OVERWORLD_MINIMAP) && has_map) ? thedmap.minimap_tile[1] : thedmap.minimap_tile[0];
+				int32_t mapcset = (!get_qr(qr_BROKEN_OVERWORLD_MINIMAP) && has_map) ? thedmap.minimap_cset[1] : thedmap.minimap_cset[0];
+				//What a mess. The map drawing is based on a variable that can change states during a scrolling transition when warping. -Z
+				if (maptile)
+				{
+					draw_block(dest, tx, ty, maptile, mapcset, 5, 3);
+				}
+				else if (c.overworld_map_tile || c.overworld_map_tile)
```
connorjclark:
```suggestion
				else if (c.overworld_map_tile)
```

src/new_subscr.cpp
```diff
@@ -3483,93 +3483,91 @@ byte SW_MMap::getType() const
...
 			{
-				if((thedmap.flags&dmfMINIMAPCOLORFIX) != 0)
+				int32_t maptile = has_map ? thedmap.minimap_tile[1] : thedmap.minimap_tile[0];
+				int32_t mapcset = has_map ? thedmap.minimap_cset[1] : thedmap.minimap_cset[0];
+				//What a mess. The map drawing is based on a variable that can change states during a scrolling transition when warping. -Z
+				if (maptile)
+				{
+					draw_block(dest, tx, ty, maptile, mapcset, 5, 3);
+				}
+				else if (c.dungeon_map_tile || c.dungeon_map_tile)
```
connorjclark:
```suggestion
				else if (c.dungeon_map_tile)
```

connorjclark (commented):
Found some issues.

Also asked Gemini to point out that bug fix we can't find. Maybe it makes sense to you: https://gemini.google.com/share/0d57c5503061

src/subscr.cpp
EmilyV99: because there is only a 1px gap vertically, so the top gap is actually covered by the bottom of the above screen- so it would end up as `if(top) y_1 -= 0;` which is a nop (doing `-= 1` would result in the region appearing to connect to the row of screens above the top screen in the region, a bug). Figured the comment was worth leaving incase some future option changes the spacing in such a way that the top would once again matter.

src/subscr.cpp
EmilyV99: oop, think deathrider's editor fucked up settings. will fix.

src/new_subscr.cpp
EmilyV99: oh lmao, these weren't new but certainly worth fixing while here lmao

src/components/zasm/table.cpp
EmilyV99: that'd be my bad oop

PR #1158, opened 2025-09-06
misc: auto-generate 'include/bindings/qrs.zh' from source code

EmilyV99:


scripts/generate_qr_constants.py
```diff
@@ -0,0 +1,47 @@
...
+    ]
+    output = subprocess.run(
+        args,
+        cwd=build_folder,
+        encoding='utf-8',
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+    if output.returncode != 0:
+        raise Exception(f'got error: {output.returncode}\n{output.stdout}')
```
connorjclark: Can replace subprocess with simpler `run_target` helper:

```
import run_target

# ...
def output_qr_enum_file():
    run_target.check_run('zeditor', [...args...], build_folder)
    # end of function
```

connorjclark (approved):

connorjclark:
error during compilation

```
../src/zq/zquest.cpp:27033:23: error: code will never be executed [-Werror,-Wunreachable-code]
                        file << (qr_MAX/8) << " * 8";
```

CMakeLists.txt
```diff
@@ -932,6 +932,16 @@ if(NOT EMSCRIPTEN AND NOT (DEFINED ENV{CI} AND CMAKE_CXX_COMPILER_ID MATCHES "Cl
...
 endif()
 
+# qr generation
+set(gen_zscript_qr_enum scripts/generate_qr_constants.py zeditor)
+add_custom_command(
+	COMMAND ${PYTHON} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/generate_qr_constants.py --build_folder "${BUILD_FOLDER}"
+	OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/resources/include/bindings/qrs.zh
+	DEPENDS ${gen_zscript_qr_enum}
+)
+add_custom_target(gen_zscript_qr_enum DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/resources/include/bindings/qrs.zh)
```
connorjclark: depends on zeditor too

connorjclark: But this will get annoying fast. it makes building zplayer require building the editor first...

I suggest removing the zplayer dep on line 943. We can build this target directly, or we can build all the targets to run it.

scripts/generate_qr_constants.py
```diff
@@ -0,0 +1,36 @@
...
+script_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+root_dir = script_dir.parent
+resource_dir = script_dir / '..' /  'resources'
+dest_file = resource_dir / 'include' / 'bindings' / 'qrs.zh'
+
+sys.path.append(str((root_dir / 'scripts').absolute()))
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    '--build_folder', help='The location of the build folder. ex: build/Release', default='build/Release'
```
connorjclark: The error in CI is because the default build folder is not right in all our environments.

```
  test:
    needs: warm-lfs-cache
    strategy:
      matrix:
        include:
          - runs-on: windows-2022
            arch: x64
            compiler: msvc
            # RelWithDebInfo invalidates ccache, and we currently can't show backtraces on crash for windows in CI
            # so no need for it yet.
            config: Release
          - runs-on: windows-2022
            arch: win32
            compiler: msvc
            config: RelWithDebInfo
            extra-args: --max_duration 60
          - runs-on: macos-13
            arch: intel
            compiler: clang
            config: RelWithDebInfo
          - runs-on: ubuntu-22.04
            arch: x64
            compiler: clang
            config: RelWithDebInfo
          - runs-on: ubuntu-22.04
            arch: x64
            compiler: gcc
            config: RelWithDebInfo
```

Try removing the default. Allow "None" to be passed, and `run_target` will then handle finding the folder itself (or use the one you provide from the CLI if given).

PR #1165, opened 2025-10-14
fix(zscript)!: @deprecated comment tag properly works on non-internal functions

EmilyV99:
Used to mark some old std_functions as deprecated


connorjclark (approved):
nice 👍

PR #1176, opened 2025-11-16
feat: 'Map Subscreens'

EmilyV99:
If a Map Subscreen is assigned to a dmap, it will replace the map button map for that dmap.


src/zasm/defines.h
```diff
@@ -3293,7 +3293,21 @@ enum ASM_DEFINE
 #define COMBOD_DIVE_UNDER_LEVEL        0x166F
 #define GAMELAYERZTHRESHOLDS           0x1670
 #define NPCFLAGS                       0x1671
-#define MESSAGEDATASEGMENTS             0x1672
+#define MESSAGEDATASEGMENTS            0x1672
+#define GAMENUMMSUB                    0x1673
+#define GAMEMSUBOPEN                   0x1674
+#define DMAPDATAFLOOR                  0x1675
```
connorjclark, line 3299: missing entry in get_register_ref_dependency for DMAPDATAFLOOR (anything using GET_REF should be in there).

EmilyV99: nice catch

PR #1182, opened 2025-11-24
Cleanup / upgrade guis

EmilyV99:


src/zq/zquest.cpp
```diff
@@ -18755,860 +18160,144 @@ int32_t onEditComboAlias()
...
     
-    if(list_size != NULL)
-        *list_size=4;
-        
-    return NULL;
+    sprintf(string, "%s", src2);
+    free(src);
+    free(src2);
+    return string;
 }
```
connorjclark: dunno why the diff is showing these functions (you didn't touch them) but this is dead code too. clean_numeric_string, strip_decimals

EmilyV99: yea its showing a few because of other nearby functions that were removed having similar content to other functions that still exist, causing it to think some removed function bodies were "moved" instead, which fucks up the whole diff.

There's definitely a lot more dead code to go through lol, and a bunch more to deadify by replacing old dialogs with new ones.

PR #1197, opened 2025-12-11
fix(zq)!: map viewer error popups being invisible

EmilyV99:
Also adds some more detail to the errors
Also allows closing the program while the map viewer is open

### DIFF
--- a/src/zq/render_map_view.cpp
+++ b/src/zq/render_map_view.cpp
@@ -142,7 +142,14 @@ void mapview_open(int flags, int sw, int sh, int bw, int bh)
 	rti_map_view->sh = sh;
 	rti_map_view->set_size(bw, bh);
 	rti_map_view->dirty = true;
-	get_root_rti()->add_child(rti_map_view);
+	
+	// Add to the back of the dialogs. This allows popups intended to show over the map to still be visible (such as the image failing to save)
+	// Adding directly to `rti_dialogs` causes `screen` to end up `nullptr` and crashes, so add to the last child of dialogs instead.
+	// Reload the tints first to ensure the last child is not `rti_tint`, otherwise the entire map vanishes on using the KEY_SPACE shortcut.
+	// `popup_zq_dialog_start()` should have been used before calling `mapview_open`, and `popup_zq_dialog_end()` should be called after `mapview_close()`.
+	reload_dialog_tint();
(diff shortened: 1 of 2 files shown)


src/zq/render_map_view.cpp
```diff
@@ -142,7 +142,14 @@ void mapview_open(int flags, int sw, int sh, int bw, int bh)
...
 	rti_map_view->dirty = true;
-	get_root_rti()->add_child(rti_map_view);
+	
+	// Add to the back of the dialogs. This allows popups intended to show over the map to still be visible (such as the image failing to save)
+	// Adding directly to `rti_dialogs` causes `screen` to end up `nullptr` and crashes, so add to the last child of dialogs instead.
+	// Reload the tints first to ensure the last child is not `rti_tint`, otherwise the entire map vanishes on using the KEY_SPACE shortcut.
+	// `popup_zq_dialog_start()` should have been used before calling `mapview_open`, and `popup_zq_dialog_end()` should be called after `mapview_close()`.
+	reload_dialog_tint();
+	ASSERT(rti_dialogs.has_children());
+	rti_dialogs.get_children().back()->add_child(rti_map_view);
```
connorjclark, line 152: would adding to the _front_ of `root_rti` address the same issues as this?

EmilyV99: No, adding it to the front means the drop-down Menu that you clicked `View Map` in will be on top of the map, since that is also in `rti_dialogs`. The map needs to be above some `rti_dialogs` children, but under other `rti_dialogs` children; so it must be inside `rti_dialogs` itself in some way.

connorjclark: makes sense. wfm.

I'll look into something a bit cleaner later ... maybe unattaching the entire render tree so that the map view is the only thing in the render graph, and re-attaching when it's closed could work well

connorjclark (approved):

src/zq/render_map_view.cpp
EmilyV99: `rti_dialogs` itself is still needed as the map view can call it's own starting dialog again when you press space - but, emptying all of `rti_dialogs` children when opening the map and then restoring them after could work (alongside the map being before `rti_dialogs`. But, personally, that sounds more messy than this solution.

PR #1205, opened 2025-12-17
fix(zc)!: bombs falling into pits not near them when exploding

EmilyV99:
this was caused the the bomb having a large hit offset when exploding, and a check was bounding the coordinates back into the valid screen area, causing the bomb to fall into pits along the right edge of the screen.

### DIFF
--- a/src/zc/maps.cpp
+++ b/src/zc/maps.cpp
@@ -865,8 +865,8 @@ bool canPermSecret(int32_t dmap, int32_t screen)
 
 int32_t MAPCOMBO(int32_t x, int32_t y)
 {
-	x = vbound(x, 0, world_w-1);
-	y = vbound(y, 0, world_h-1);
+	if (!is_in_world_bounds(x, y))
+		return 0;
 	int pos = COMBOPOS(x%256, y%176);
 	mapscr* scr = get_scr_for_world_xy(x, y);
 	return scr->data[pos];
--- a/src/zc/guys.cpp
+++ b/src/zc/guys.cpp
@@ -8445,6 +8445,16 @@ eTektite::eTektite(zfix X,zfix Y,int32_t Id,int32_t Clk) : enemy(X,Y,Id,Clk)
 	if (SIZEflags != 0) init_size_flags();;
 }
 
+static bool tek_combo_check(zfix tx, zfix ty)
+{
+	// tektites treat the world edge as a no-go zone if the combo next to the edge is a no-go zone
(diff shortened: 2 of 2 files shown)


EmilyV99:
`MAPCOMBO` (and `MAPCOMBOTYPE` by extension) were missing a check for being in-bounds of the region- while `MAPFLAG`, `MAPCSET`, `MAPCOMBOL`, `MAPFLAGL`, `MAPCSETL`, `MAPCOMBOFLAGL`, `MAPCOMBOFLAG`, etc have the check.

Tektites in specific seemed to have a change with fixing this, related to no-jump and no-enemy combo types (via `MAPCOMBOTYPE`) - in this case, I think keeping them bounded to catch types they are flying past makes the most sense, so that's how I've cleaned up the replay failure (I hope, it's still running rn)

src/zc/maps.cpp
```diff
@@ -865,6 +865,8 @@ bool canPermSecret(int32_t dmap, int32_t screen)
 
 int32_t MAPCOMBO(int32_t x, int32_t y)
 {
+	if (!is_in_world_bounds(x, y))
+		return 0;
 	x = vbound(x, 0, world_w-1);
 	y = vbound(y, 0, world_h-1);
```
connorjclark: can remove the vbounds now

connorjclark:
> fix(zc)!: bombs falling into pits not near them when exploding #1205

remove the `!`? unless this introduced in a commit that is only in 3.0 and this doesn't need to be cherry-picked to 2.55.

connorjclark (approved):

connorjclark:
great catch with the Tektites stuff. I guess that was why I skipped doing this for MAPCOMBO initially.

EmilyV99:
> > fix(zc)!: bombs falling into pits not near them when exploding #1205
> 
> remove the `!`? unless this introduced in a commit that is only in 3.0 and this doesn't need to be cherry-picked to 2.55.

Ah, the QR in question (that caused the issue with the bombs) is in 2.55, so it will need cherry-pick, good catch.

EmilyV99:
> great catch with the Tektites stuff. I guess that was why I skipped doing this for MAPCOMBO initially.

Yeah, I just made the change and went "well let's see how many things break tremendously"... only for there to be only 3 failures, all involving tektites, at the edge of the screen, with NES dungeon wall combos involved. Easily traced to the MAPCOMBOTYPE checks from there. . . . and the simple fix just worked.

PR #1211, opened 2025-12-19
fix(zc): getting stuck on solid FFCs w/ new movement disabled

EmilyV99:


ZoriaRPG:
This still exists Good job, I suppose.

PR #1213, opened 2025-12-21
Control Schemes: Multiple sets of keybinds, changable on a per-quest basis

EmilyV99:
https://discord.com/channels/876899628556091432/1281775489618546780


src/zc/scripting/types/input.cpp
```diff
@@ -186,43 +186,12 @@ static ArrayRegistrar KEYBINDINGS_registrar(KEYBINDINGS, []{
...
-				case 8: Rkey = value; break;
-				case 9: Pkey = value; break;
-				case 10: Exkey1 = value; break;
-				case 11: Exkey2 = value; break;
-				case 12: Exkey3 = value; break;
-				case 13: Exkey4 = value; break;
-				default: NOTREACHED();
-			}
-
+			// read-only, scripts can't change the user's keybinds
```
connorjclark, line 194: call `impl.setReadOnly();` below.

src/base/render.cpp
```diff
@@ -804,6 +804,7 @@ void popup_zqdialog_start_a5()
 	set_bitmap_create_flags(true);
 	rti->bitmap = create_a5_bitmap(zq_screen_w, zq_screen_h);
 	al_set_new_bitmap_flags(0);
+	rti->set_transform({ 0, 0 });
```
connorjclark: this fixes one instance, not the general issue. RenderTreeItem constructor should set the default transform.

src/dialog/text_prompt.h
connorjclark, line 1: This isn't "picking" text, it's inputing or prompting for text. please rename to text_input.h or text_prompt.h + add a short comment explaining what it does somewhere.

commit suggestion: `feat!: add dialog for inputing text` (on it's own it isn't of interest to add to the changelog, since it's only used in a new dialog)

src/zc/control_scheme.h
connorjclark, line 1: If what is in base_config/zc.cfg is no longer utilizing `[Controls]` for a fresh install, it should be removed.

Additionally, for discoverability an (empty) `base_config/controls.cfg` would be useful, especially if it had comments showing what the default mappings are.

src/zc/control_scheme.cpp
```diff
@@ -0,0 +1,392 @@
...
+	cheatkeys[Cheat::ShowFFCScripts][0] = KEY_F;
+	
+	memset(stick_data, 0, sizeof(stick_data));
+	
+	// set y axis to second axis
+	stick_data[control_scheme::stick_1][control_scheme::axis_y][control_scheme::data_axis] = 1;
+	stick_data[control_scheme::stick_1][control_scheme::axis_y][control_scheme::data_axis] = 1;
+	// set second stick to second stick
+	stick_data[control_scheme::stick_1][control_scheme::axis_x][control_scheme::data_stick] = 1;
+	stick_data[control_scheme::stick_1][control_scheme::axis_y][control_scheme::data_stick] = 1;
```
connorjclark: this says stick_1, not 2.

src/dialog/control_schemes.cpp
```diff
@@ -0,0 +1,242 @@
+#include "control_schemes.h"
```
connorjclark, line 2: can you explain the design at the top of this file?

for instance, I didn't know until far down in `refresh_edit_buttons` that the "default" mapping is not editable. It also took awhile to realize that you are importing the old control settings as the global scheme when first run.

While it's fresh, a couple lines at the top of this file describing the thing you just designed would be useful long term.

src/zconfig.cpp
```diff
@@ -192,3 +193,229 @@ bool is_feature_enabled(const char* flag_switch, const char* config_header, cons
...
+{
+	if (!data) return default_val;
+	const char* str = al_get_config_value(data, header, name);
+	if (!str) return default_val;
+	return std::stod(str);
+}
+char const* zc_a5_cfg::get_config(char const* header, char const* name, char const* default_val)
+{
+	if (!data) return default_val;
+	const char* str = al_get_config_value(data, header, name);
```
connorjclark, line 322: The `const char*` returned here is owned by `ALLEGRO_CONFIG`, which means it gets destroyed whenever the config is reloaded. That's a footgun - we should just return a std::string here to be safe.

src/zc/control_scheme.cpp
```diff
@@ -0,0 +1,392 @@
...
+		control_scheme scheme;
+		scheme.load_from_old_section(ctrl_sect);
+		control_config.add_comment("Custom", "Auto-generated from old config file");
+		scheme.save_to_section("Custom");
+		control_schemes["Custom"] = scheme;
+		global_control_scheme_name = "Custom";
+		save_schemes();
+	}
+	control_config.remove_section(DEFAULT_CONTROL_SCHEME_NAME);
+	control_config.add_comment(DEFAULT_CONTROL_SCHEME_NAME, "Non-modifiable. Changes are automatically reset.");
```
connorjclark, line 140: Changes from where are reset? is the global scheme written to the cfg but just ignored when loading?

connorjclark (commented):
I've not yet tested with my controllers. Please wait on merging until I can do that.

src/dialog/text_prompt.h
EmilyV99: It's named "TextPick" to match the existing "NumPick"... should I rename both?

You are "picking" what you want to name something, so the name works fine imo...

src/zc/control_scheme.h
EmilyV99: What's in [Controls] there is effectively being used to create `controls.cfg` on a fresh install... and the a5 config stuff I set up isn't set up to handle the base_config stuff

src/zc/control_scheme.cpp
EmilyV99: On loading, the default config is overwritten, yes. I could also stop it from writing out to the file, but it seemed worthwhile to have the values there for people to copy from (which, kinda acts as a replacement for `base_config` not being hooked up to this)

So, when it loads the file, it outright deletes any scheme named `Default`, and then it replaces it with a fresh one that is exactly the default settings. This does not affect what scheme you have "set"- unless of course it was the Default one, in which case any changes to it would be reset, but you can't change it in the editor anyway. So, manually editing the default in `controls.cfg` is the only place you could edit it, thus why I put the config comment there to tell you not to.

The "Global Scheme" is just a single string naming which scheme is set globally- that is saved as part of `zc.cfg`'s `[Controls]` section still. The actual controls FOR that scheme are in the config section *of that scheme name* in `controls.cfg`.

src/zc/control_scheme.cpp
EmilyV99: ACK, nice catch.

src/base/render.cpp
EmilyV99: ah, fair.

src/dialog/control_schemes.cpp
EmilyV99: will do (though belongs more in `zc/control_scheme.cpp` than `dialogs/control_schemes.cpp`)

src/zconfig.cpp
EmilyV99: believe nullptr is valid so would want to be `optional<string>` then?

src/dialog/text_prompt.h
connorjclark: It wasn't obvious from me what it does without reading the code closely. Keep the name if you feel strongly about changing it. imo: the name may match something that exists in the code already but that doesn't mean it's the best name.

At least leave a comment describing that it is a free-form text input.

> You are "picking" what you want to name something, so the name works fine imo...

A "picker" dialog implies from a constrained set of choices, sometimes but not always through an interactive widget (like a color wheel for a color picker). In no UX framework I've heard of would a free form text input be described as a "picker", which is why I didn't know what this dialog does from just the name.

src/zconfig.cpp
connorjclark: Sure. In that case "default" should not be part of the function interface, but rather deferred to the caller via optional (`.value_or`).

src/dialog/text_prompt.h
EmilyV99: text_prompt it is

src/base/render.cpp
EmilyV99: should be fixed now I think, I just added `= 0;`s in the class definition for Transform and it appears to be working.

src/zconfig.cpp
EmilyV99: K, reworked it all to use optionals with `.value_or()`. Had to change the getter names, because the type of the default val was the only thing overloading them... but that's worthwhile.

src/zc/control_scheme.h
EmilyV99: base config is cleaned up. The defaults will be available for reference as soon as zplayer is launched for the first time, as the `Default` scheme is written out to the `controls.cfg` file for reference.

connorjclark: sound good

src/zc/zc_sys.cpp
```diff
@@ -6740,7 +6740,7 @@ void System()
 	the_player_menu.position(0, 0);
 	
 	bool running = true;
-	bool esc = key[KEY_ESC] || cMbtn();
```
connorjclark, line 7744: Although this lets the controller open the system menu, no controller input lets the user navigate or (most importantly) close it. I don't want to get hung up on making the menu actually work w/ a controller right now, but ideally it shouldn't be possible to get stuck (which means possibly needing to get off the couch and walk back to the keyboard).

src/dialog/control_binding.h
```diff
@@ -0,0 +1,59 @@
...
+	
+	GUI::ListData gamepad_list;
+	
+	void load_kb_labels();
+	void load_gamepad_labels();
+	void load_cheat_labels();
+	
+	void check_joystick();
+	
+	bool check_no_duplicate_keys() const;
```
connorjclark, line 59: When I make a new scheme, edit it, then press OK w/o any modifications I see this:

image

connorjclark: image

EmilyV99: yeeep, `LControl + RControl` -> `Command + Command` becomes a conflict on mac specifically. Simple fix.

src/zc/zc_sys.cpp
EmilyV99: The intent was for the menu button to also close the system menu, I think the issue is just a missing `poll_joystick()` call. See if it's fixed with this push.

connorjclark (approved):

PR #1226, opened 2025-12-30
fix(zc): `Disabled In Sideview Water` not working on some item types

EmilyV99:


Alucard648:
I am guessing, it will be possible to refactor "useable as bunny" flag in the same way.

EmilyV99:
> I am guessing, it will be possible to refactor "useable as bunny" flag in the same way.

The usable as bunny flag has an error sound it can be set to play, so, no- because it can't do that if it's as though the button was never pressed in the first place.

PR #1233, opened 2026-01-17
Remove `Attribytes` and `Attrishorts`, inrease `Attributes` size

EmilyV99:


src/zc/hero.cpp
```diff
@@ -8113,7 +8113,7 @@ heroanimate_skip_liftwpn:;
 							decorations.add(new dRipples(x, y, dRIPPLES, 0, ripplesprite));
 						if (watercmb.usrflags&cflag2)
 						{
-							if (auto ilvl = current_item(watercmb.c_attributes[10].getTrunc()); !(ilvl > 0 && ilvl >= watercmb.c_attributes[11].getTrunc()))
+							if (auto itype = watercmb.c_attributes[10].getTrunc(); !itype || current_item(itype) >= watercmb.c_attributes[11].getTrunc())
```
connorjclark: > current_item(itype) >= watercmb.c_attributes[11].getTrunc()

you're sure the conditional `>=` is the right direction here? something being a higher level should prevent damage shouldn't it?

src/gui/scrolling_pane.cpp
```diff
@@ -118,75 +136,409 @@ int32_t scrollProc(int32_t msg, DIALOG* d, int32_t c)
...
+					}
+					else if (mx >= d->x + tw - 2 - bh)
+					{
+						clicked_on = bottom_btn;
+						tx = d->x + tw - 2 - bh;
+					}
+					else if (tw <= 32 + 6)
+						clicked_on = bar;
+					else if (mx < d->x + 2 + bh + pos)
+						clicked_on = top_bar;
```
connorjclark: "top" and "bottom" here are being used for horizontal right and left? pretty confusing. different names for these enums would help.

resources/include/bindings/combodata.zh
connorjclark, line 1: `refactor(zscript):` or `feat(zscript):` ?

src/zc/hero.cpp
EmilyV99: Nice catch, yeah, that seems backwards (double negatives are confusing...)

resources/include/bindings/combodata.zh
EmilyV99: Changing the size of the attribute fields directly affects engine stuff too, such as Armos / Grave combos which had an enemy ID wrongfully stored in attribytes before, can now use enemies >255.

resources/include/bindings/combodata.zh
```diff
@@ -202,7 +202,7 @@ class combodata {
 	// @zasm_var COMBODANIMFLAGS
 	internal int AnimFlags;
 
-	// The 4 combo attributes. These have a full range (all zscript values,
+	// The 24 combo attributes. These have a full range (all zscript values,
 	// including decimal). Their effect depends on the combo type.
 	//
 	// @zasm_var COMBODATTRIBUTES
```
connorjclark, line 209: @versionchanged ?  so 2.55 users know the length is 4.

PR #1247, opened 2026-02-10
Increase Max Items

EmilyV99:


src/base/qst.cpp
```diff
@@ -1151,7 +1009,7 @@ bool reset_items(bool validate, zquestheader *Header)
     if (get_app_id() == App::zquest)
         ret = init_section(Header, ID_ITEMS, NULL, NULL, validate, "modules/classic/default.qst");
     
-    for(int32_t i=0; i<MAXITEMS; i++) reset_itemname(i);
```
connorjclark, line 1154: If this is on purpose, leave a comment for why MAXITEMS is not appropriate here.

src/base/qst.cpp
```diff
@@ -8619,1036 +8225,1531 @@ int32_t readitems(PACKFILE *f, word version, word build)
...
+		for(int32_t i=0; i<256; i++)
+			reset_itemname(i);
+    }
+    
+    for (int32_t i=0; i<items_to_read; i++)
+		if (auto ret = read_single_item_old(f, s_version, i, version, build))
+			return ret;
+
+	if (!should_skip)
+		for(word i = 0; i < itemsbuf.capacity(); ++i)
```
connorjclark, line 9711: Why is _capacity_ of interest? Why does how much memory is pre-allocated matter? This seems like it should be `size()`.

If for some reason this was on purpose, something else probably needs to change to make the code be more reasonable... (call .resize somewhere at the start, so the `.size()` is actually the length of valid items in the container?)

src/base/qst.cpp
```diff
@@ -8619,1036 +8225,1531 @@ int32_t readitems(PACKFILE *f, word version, word build)
...
+		itemsbuf.clear();
+		for (int q = items_to_read; q < iLast; ++q)
 		{
-			switch (tempitem.type)
-			{
-				case itype_sword:
-					tempitem.usesound2 = WAV_BEAM;
-					break;
-			}
+			itemdata& id = itemsbuf[q];
```
connorjclark, line 9682: This relies on bounded_vec's non-standard behavior of auto-resizing on access (no std container does this). Aside from being confusing to read (since clear() usually means the container is empty), this loop will cause the underlying vector to reallocate memory on every single iteration (see `bounded_vec::at`, which re-allocates the entire array by growing it one item at a time).

To make less confusing and more performant, can replace `clear()` with `.resize(whatever the max item id is)` - then that will avoid all dynamic resizing on `operator[]`.

src/base/qst.cpp
```diff
@@ -8619,1036 +8225,1531 @@ int32_t readitems(PACKFILE *f, word version, word build)
...
+        {
+            return qe_invalid;
+        }
+    }
+    
+    
+	if (!should_skip)
+	{
+		itemsbuf.clear();
+		for (int q = items_to_read; q < iLast; ++q)
```
connorjclark, line 9680: Is this supposed to not start at 0?

connorjclark (commented):
Uploaded replays pass.

src/base/qst.cpp
EmilyV99: ah, yeah, should resize it before duh.

src/base/qst.cpp
EmilyV99: `capacity()` is the size of the underlying vector- so, any element `>= capacity()` will return a default element.
`.size()` for bounded_vec is not the size of the underlying vector- it is the *accessible size*, including the default elements at the end.

src/base/qst.cpp
EmilyV99: This is specifically defaulting things that won't be hit in the following loop because their ID is too high, if those indexes have `i` constants meaning they may be needed for hardcoded stuff (`< iLast`).

src/base/qst.cpp
EmilyV99: actually removing that loop is probably more correct

PR #1262, opened 2026-03-12
feat: expand max Sprite Data (256 -> 8192)

EmilyV99:
Done:
- Changed sprite data in memory
- renamed from `wpndata` -> `sprite_data`, `wpnsbuf` -> `sprite_data_buf`.
- Increase sprite ID storage vars `byte`->`word`

### DIFF
--- a/src/dialog/info_lister.cpp
+++ b/src/dialog/info_lister.cpp
@@ -24,8 +24,6 @@
 #include <sstream>
 #include "advanced_music.h"
 
-extern char *weapon_string[];
-
 static string get_info(bool sel, bool advpaste, bool saveload = true, bool copypaste = true)
 {
 	std::ostringstream oss;
@@ -275,7 +273,7 @@ void ItemListerDialog::preinit()
 		if (itemsbuf.capacity() < MAXITEMS)
 		{
 			frozen_end = 1;
-			lister.add(fmt::format("<New Item> ({:03})", itemsbuf.capacity()), itemsbuf.capacity());
+			lister.add(fmt::format("<New Item> ({:04})", itemsbuf.capacity()), itemsbuf.capacity());
 		}
 		resort();
 		if(selected_val < 0)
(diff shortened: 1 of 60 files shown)


src/dialog/info_lister.cpp
```diff
@@ -614,26 +651,63 @@ void SpriteListerDialog::edit()
...
+}
 void SpriteListerDialog::copy()
 {
 	copied_sprite_id = selected_val;
 	update();
 }
 bool SpriteListerDialog::paste()
 {
-	if(copied_sprite_id < 0 || selected_val < 0)
+	if (unsigned(copied_item_id) >= sprite_data_buf.capacity())
```
connorjclark: item id? what happened to sprite id? here and within rest of function

src/items.cpp
```diff
@@ -769,9 +780,106 @@ void draw_lens_hint_item(BITMAP *dest,int32_t x,int32_t y,int32_t item_id, int32
...
+#ifdef IS_PLAYER
+	// In all honesty, these look buggy AF if more than one of the same
+	// is on-screen, because each one is one frame advanced in the animation
+	// from the previous. Should ideally overhaul these hints to update animation
+	// globally instead of during each draw, so the animations don't do this.
+	// Out of scope for current changes, so leaving for future fixing. -Em
+	LensHintData data = lens_hint_sprite.get(sprite_id);
+	if (weapon_id > -1)
+	{
+		auto const& bugged_offset = lens_hint_weapon_offsets.get(sprite_id);
```
connorjclark: bugged_offset ? not used.

EmilyV99: nice catch, leftover from something I thought *might* need a replay check that ended up not needing one

src/dialog/info_lister.cpp
EmilyV99: oop! copy/paste error strikes again, clearly missed testing that in the gui

PR #1263, opened 2026-03-14
feat(zscript): Falsy-Coalescing (?:, ?:=) operator

EmilyV99:
`a ?: b` == `a ? a : b`
`a ?:= b` == `unless (a) a = b;`


src/parser/BuildVisitors.cpp
```diff
@@ -3566,6 +3566,61 @@ void BuildOpcodes::caseExprRShift(ASTExprRShift& host, void* param)
...
+	{
+		CONST_VAL(*v);
+		return;
+	}
+	auto lval = host.left->getCompileTimeValue(this, scope);
+	if(lval)
+	{
+		if(*lval)
+			visit(host.right.get(), param);
+		else DCHECK(false); // should've exited via getCompileTimeValue()
```
connorjclark, line 3581: hm? can't lval be a constant 0?

EmilyV99: ...yep, backwards. Needed to be `!*lval`, because it can't be constant NON-zero- constant non-zero would have exited above due to the whole expression being constant, so this section should then be unreachable. Just had the condition flipped- nice catch.

PR #1266, opened 2026-03-15
Cleanup bounded container

EmilyV99:
the first and last commit change stuff about bounded containers in general; all the middle commits are just updating a bunch of `[]` access to `.get()` access for various containers.

### DIFF
--- a/src/base/containers.h
+++ b/src/base/containers.h
@@ -159,6 +159,7 @@ public:
 	
 	virtual value_type& at(size_type ind) = 0;
 	virtual value_type const& at(size_type ind) const = 0;
+	virtual value_type const& get(size_type ind) const {return at(ind);}
 	virtual value_type& operator[](size_type ind) = 0;
 	virtual value_type const& operator[](size_type ind) const = 0;
 	virtual void clear() = 0;
@@ -218,10 +219,6 @@ public:
 			return this->default_val;
 		}
 		throw std::out_of_range("Bad bounded_map access");
-	}
-	value_type const& get(size_type ind) const
-	{
-		return at(ind);
 	}
 
 	value_type& operator[](size_type ind)
@@ -288,13 +285,13 @@ public:
 			return false;
 		if(other.default_val != this->default_val)
 			return false;
-		map<size_type,bool> keys;
-		for(auto [k,v] : cont)
-			keys[k] = true;
+		std::set<size_type> keys;
+		for(auto [k,v] : this->cont)
+			keys.insert(k);
 		for(auto [k,v] : other.cont)
-			keys[k] = true;
-		for(auto [k,b] : keys)
-			if(other[k] != (*this)[k])
+			keys.insert(k);
+		for(auto k : keys)
+			if(other.get(k) != this->get(k))
 				return false;
 		return true;
 	}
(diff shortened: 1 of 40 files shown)


src/base/containers.h
```diff
@@ -184,7 +189,15 @@ class bounded_container
...
-template<typename Sz,typename T>
+template<typename ContType, typename ElemType>
+concept is_bounded_container = (
+	std::is_base_of<bounded_container<uint8_t, ElemType>, ContType>::value
+	|| std::is_base_of<bounded_container<uint16_t, ElemType>, ContType>::value
+	|| std::is_base_of<bounded_container<uint32_t, ElemType>, ContType>::value
+	|| std::is_base_of<bounded_container<uint64_t, ElemType>, ContType>::value
+);
+
+template<uint_type Sz,typename T>
```
connorjclark, line 198: unsure how portable this is (size_t may not be captured here on all platforms, perhaps signed types don't resolve, idk).

here are two other options.

first is equivalent but checks for any numeric Sz. it also extracts the concepts to a readable namespace+template.

```cpp
namespace detail
{
	template <std::unsigned_integral Sz, typename T>
	std::true_type test_bounded_container(const bounded_container<Sz, T>*);

	std::false_type test_bounded_container(const void*);
}

template <typename ContType>
concept is_bounded_container = decltype(detail::test_bounded_container(std::declval<ContType*>()))::value;
```

or just test if `.get(idx)` resolves:

```cpp
template <typename ContType>
concept is_bounded_container = requires(const ContType& cont, size_t idx)
{
	{ cont.get(idx) };
};
```

src/zalleg/packfile.h
```diff
@@ -679,7 +679,7 @@ inline bool p_putbvec(bounded_vec<Sz,T> const& cont, PACKFILE *f)
 		{
 			for(Sz q = 0; q < writecnt_v; ++q)
 			{
-				if(cont[q] == dt) continue;
+				if(cont.get(q) == dt) continue;
 				if(!p_putvar(q, f))
 					return false;
 				if(!p_putvar(cont[q], f))
```
connorjclark: missed one (but iiuc it doesn't matter for this function, since this is a const reference so there must be no allocation)

src/base/containers.h
```diff
@@ -1,9 +1,13 @@
...
+#pragma once
 
 #include "base/headers.h"
 #include <array>
 #include <stdexcept>
+#include <type_traits>
+#include <concepts>
+
+template<typename T>
+concept uint_type = (std::is_integral<T>::value && std::is_unsigned<T>::value);
```
connorjclark: i think this is just `std::unsigned_integral<T>`

connorjclark:
I did minimal runtime testing - I only confirmed that the time to load yuurand.qst did not noticeably change, and that the uploaded replays still pass.

src/zalleg/packfile.h
EmilyV99: yeah the const references I updated for consistency not out of necessity

src/base/containers.h
EmilyV99: this is definitely where I most expected a suggestion from you, and as expected you have delivered! Just testing that `.get(idx)` resolves should be sufficient for the use case, though I'll definitely rename the concept then to match.

PR #1268, opened 2026-03-16
feat(zscript): ffc->ScreenFFCIndex

EmilyV99:


resources/include/bindings/ffc.zh
```diff
@@ -52,6 +52,15 @@ class ffc {
 	//
 	// @zasm_var FFCID
 	internal const int ID;
+	
+	// The FFC's index within the current screen (1-indexed)
+	// The general formula for this index is:
+	// > (index into screen's ffc array) + 1
+	// or, relative to the ID,
+	// > ((ID - 1) % 128) + 1
```
connorjclark: does this look right in the tooltip? you may need a new line, or just wrap it in backticks rather than using a block quote

PR #1278, opened 2026-03-27
refactor(zscript): remove multiple compiler compat settings

connorjclark:
The following compiler compat settings have been removed in order to
simplify the compiler. They don't seem to be necessary to support
any popular scripts, and sometimes resulted in bugs when users
inadvertently had them enabled.

* 2.50 division truncation
* 2.50 value for 'true' is 0.0001
* Binary operations use true 32-bit int
* True MAX_INT sizing (smaller max int sizes no longer supported)
* Short-circuit boolean operations (non-short circuiting no longer supported)


src/parser/BuildVisitors.cpp
```diff
@@ -2868,61 +2733,31 @@ void BuildOpcodes::caseExprOr(ASTExprOr& host, void* param)
...
-		visit(host.left.get(), param);
-		//Check left, skip if true
-		addOpcode(new OCompareImmediate(new VarArgument(EXP1), new LiteralArgument(0)));
-		addOpcode(new OGotoFalseImmediate(new LabelArgument(skip)));
-		//Get rightx
-		//Get right
-		visit(host.right.get(), param);
-		addOpcode(new OCompareImmediate(new VarArgument(EXP1), new LiteralArgument(0)));
-		//Set output
-		Opcode* ocode = new OSetCompare(new VarArgument(EXP1), new CompareArgument(CMP_NE | (decret?CMP_SETI:0)));
```
connorjclark, line 2897: was `decret?CMP_SETI:0` right before? figured it should have been `decret?0:CMP_SETI` ...

EmilyV99: that looks like it was probably wrong before

PR #1283, opened 2026-04-18
fix(zc): 'Enemies->Stay Dead' preventing 'Enemy Carries Item' items from spawning

EmilyV99:


EmilyV99:
wait I forgot to compile, my test that I thought was testing the fix actually didn't have the fix but had no issue. So uh, actually failed to reproduce the bug. oops.

PR #1300, opened 2026-04-25
Fix current_screen value

EmilyV99:
(the large removals in `setX()` and `setY()` are just removing code duplication by calling a basically identical function instead of having entirely duplicate code)

### DIFF
--- a/src/zc/maps.cpp
+++ b/src/zc/maps.cpp
@@ -3361,6 +3361,8 @@ void update_freeform_combos()
 		for_every_ffc([&](const ffc_handle_t& ffc_handle) {
 			mapscr* scr = ffc_handle.scr;
 			ffcdata& thisffc = *ffc_handle.ffc;
+			
+			thisffc.update_current_screen();
 
 			// Combo 0?
 			if(thisffc.data==0)
@@ -3527,6 +3529,7 @@ void update_freeform_combos()
 				}
 			}
 			thisffc.solid_update();
+			thisffc.update_current_screen();
 		});
 	}
 }
@@ -6163,7 +6166,7 @@ static void load_a_screen_and_layers_init(int screen, bool screen_overlay, bool 
 			if ((previous_scr->ffcs[i].flags&ffc_carryover))
 			{
 				auto& ffc = base_scr->getFFC(i) = previous_scr->ffcs[i];
-				ffc.screen_spawned = screen;
+				ffc.screen_spawned = ffc.current_screen = screen;
 
 				ffc_id_t ffc_id = get_region_screen_offset(screen)*MAXFFCS + i;
 				loadscr_ffc_script_ids_to_remove.erase(ffc_id);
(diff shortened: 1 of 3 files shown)


src/zc/maps.cpp
```diff
@@ -6163,7 +6166,7 @@ static void load_a_screen_and_layers_init(int screen, bool screen_overlay, bool
 			if ((previous_scr->ffcs[i].flags&ffc_carryover))
 			{
 				auto& ffc = base_scr->getFFC(i) = previous_scr->ffcs[i];
-				ffc.screen_spawned = screen;
+				ffc.screen_spawned = ffc.current_screen = screen;
```
connorjclark, line 6169: good catch

src/zc/maps.cpp
```diff
@@ -3361,6 +3361,8 @@ void update_freeform_combos()
 		for_every_ffc([&](const ffc_handle_t& ffc_handle) {
 			mapscr* scr = ffc_handle.scr;
 			ffcdata& thisffc = *ffc_handle.ffc;
+			
+			thisffc.update_current_screen();
```
connorjclark, line 3365: this and the 2 changes below are the only changes i'm comfortable making without spending a lot more time on this

src/zc/maps.cpp
```diff
@@ -6179,9 +6182,10 @@ static void load_a_screen_and_layers_init(int screen, bool screen_overlay, bool
 	int num_ffcs = base_scr->numFFC();
 	for (word i = 0; i < num_ffcs; i++)
 	{
-		base_scr->ffcs[i].screen_spawned = screen;
-		base_scr->ffcs[i].x += offx;
-		base_scr->ffcs[i].y += offy;
+		auto& ffc = base_scr->ffcs[i];
+		ffc.screen_spawned = ffc.current_screen = screen;
```
connorjclark, line 6186: mapscr getFFC/shrinkToFitFFCs already sets current_screen, but i guess this doesn't hurt 🤷

it was an issue in the above block just b/c the ffc was copied from a different screen. but otherwise, mapscr's initialization ensures the current_screen is set correctly.

EmilyV99: but those only apply if the number of ffcs has CHANGED, if the number of ffcs is the same for shrinkTofitFFCs or if the ffc index being "gotten" by getFFC is in-bounds of the current vector, then the code for updating current screen there ***does not run***- that's the whole bug Moosh was reporting

connorjclark: Did you confirm this in a debugger? Here's how I expect it to be working:

* Canonical screens, when setup in `readmapscreen`, calls `shrinkToFitFFCs`. That should correctly set everything to the right values
* In `load_a_screen_and_layers_init` a new mapscr is copied from a canonical screen, so it should have all the values set already

It's important that the other stuff initializing the ffc inside mapscr (not just current_screen) is running, so if it's not we need to understand why better.

I wish the ffc allocation code were simpler to reason about...

----

It's fine to keep this. But the code actually isn't quite right. it should also call `update_current_screen()` which sets it based on the hit center... right now it might get update to a different value on the next frame without the ffc even moving.

EmilyV99: I didn't actually confirm it in the debugger, but Moosh's bug of it just returning `0` would not occur if it were initialized properly at all at any point. He was expecting the spawn screen value, but got `0` instead.

connorjclark: found the issue: when  `readmapscreen` calls `update_current_screen` via `shrinkToFitFFCs`, it changes from the correct value to 0. The reason is that `update_current_screen` just doesn't work outside the game (this is during loading, so no current map/screen stuff is set and `get_screen_for_world_xy` returns 0).

PR #1325, opened 2026-05-10
Fix darkroom map stuff

EmilyV99:


src/zc/maps.cpp
```diff
@@ -8068,34 +8100,120 @@ void ViewMap()
...
+				dither_offy = -yy;
+				int dx = x * 256;
+				int dy = y * 176;
+				if (scr->flags & fDARK)
 				{
-					draw_trans_sprite(screen_bmp, darkscr_bmp, 0, 0);
-					if (get_qr(qr_NEWDARK_TRANS_STACKING))
-						draw_trans_sprite(screen_bmp, darkscr_bmp_trans, 0, 0);
+					int rx = x, ry = y, rw = 1, rh = 1;
+					if (int rid = get_region_id(cur_map, screen))
```
connorjclark, line 8112: calculate_region does this for you

EmilyV99: > calculate_region does this for you

oop. so it does. not changing that now since this works (and contest deadline approaching), can clean it up later if we care

PR #1342, opened 2026-05-15
fix(zq): use actual qst file code for export formats

connorjclark:
The following export formats in the editor no longer use code separate from the main qst file code, which means these will no longer fall out of date as the qst format continues to evolve: .znpc, .ztileset, .zdoors, .zcombo


src/zq/zquest.cpp
```diff
@@ -812,7 +840,7 @@ static NewMenu export_graphics
 	{ "&Combo Set", onExport_Combos },
 	{ "Combo Pack", onExport_Combopack },
 	{},
-	{ "Combo &Alias Pack", onExport_Comboaliaspack },
+	{ "&Combo Alias", onExport_ComboAlias },
```
EmilyV99:
```suggestion
	{ "Combo &Alias", onExport_ComboAlias },
```
The &C here won't work since that's already used by `&Combo Set` above

src/zq/zquest.cpp
```diff
@@ -768,8 +806,7 @@ static NewMenu import_graphics
 	{ "Combo Pack (Full, 1:1)", onImport_Combopack },
 	{ "Combo Pack to... (Dest)", onImport_Combopack_To },
 	{},
-	{ "Combo &Alias Pack", onImport_Comboaliaspack },
-	{ "Combo A&lias Pack to...", onImport_Comboaliaspack_To },
+	{ "&Combo Alias", onImport_ComboAlias },
```
EmilyV99:
```suggestion
	{ "Combo &Alias", onImport_ComboAlias },
```
The &C here won't work since that's already used by `&Combo Set (Range)` above

EmilyV99 (dismissed):
skimmed through most of it, tested the tileset export/import fine. Mostly looks ok, and having proper testing of this helps ensure no issues, I trust that you've done this correctly.
Just one tiny GUI issue (which appears to have been present before in the '2.50 broken' menu, but now is in the main import/export menus) which is a dead simple fix (the `&` in these menus underlines the next character *and* uses that key as a key shortcut- having multiple of the same underlined letter in the same menu means the lower one's keybind will never work)

PR #1351, opened 2026-05-18
bunch of misc stuff

EmilyV99:


resources/include/bindings/hero.zh
```diff
@@ -246,6 +246,8 @@ class hero {
 	// @zasm_var HEROHEALTHBEEP
 	internal int HealthBeep;
 
+	// If the player is invisible or not.
+	// @deprecated Use [InvisTimer] instead!
```
connorjclark: maybe just reference the other field, not deprecate, since we won't backport this and there are ~20 scripts that use Invisible

resources/include/bindings/sprite.zh
```diff
@@ -137,6 +137,14 @@ class sprite {
 	// @zasm_var SPRITE_SCRIPT_TILE
 	internal int ScriptTile;
 
+	// The sprite is invisible for this many frames.
+	// If < -1, invisibility does not run out. If == -1, this is
+	// "engine invisibility", and the engine might clear it at any time.
+	//
+	// @versionadded 3.0
+	// @zasm_var SPRITE_INVIS_CLK
+	internal int InvisTimer;
```
connorjclark: saving a few characters for a variable that is isn't long in the first place is usually a mistake. readability > terseness

resources/include/bindings/hero.zh
EmilyV99: I mean, I don't see that really as a reason not to deprecate? there are definitely scripts that use a LOT of stuff that's deprecated already . . .

resources/include/bindings/sprite.zh
EmilyV99: so, `InvisibleTimer` instead? fair

resources/include/bindings/hero.zh
connorjclark: yes but i am trying to respond to community feedback and without a meaningful reason to deprecate, think maybe we should stop marking things that way

EmilyV99: I mean, deprecation does nothing under default settings, you have to opt in to caring about it. Orithan is the only one I know who was actually arguing against it fervently, and they seem not to even understand the concept fully to the point where they weren't saying the word `deprecate` at all in the conversation.

PR #1355, opened 2026-05-20
feat(zc): improve map view screen state and trigger handling

connorjclark:
- The map view now correctly applies all screen state changes, including switches and screen load triggers.
- Added new trigger conditions: "Map View Only" and "Map View Never", allowing triggers to run exclusively in the map view or be excluded from it.

Agent: https://gist.github.com/connorjclark/07737ced393b9dfe47d6517c46b118f9
Discord: https://discord.com/channels/876899628556091432/1503679477572571146


src/zc/maps.cpp
```diff
@@ -1253,21 +1267,19 @@ static void apply_screen_state_remove_combos(const screen_handles_t& screen_hand
...
 		bool from_active_screen = false;
-		trigger_secrets_for_screen_internal(screen_handles, do_layers, from_active_screen, -3, secrets_do_replay_comment);
+		bool high16only = false;
+		trigger_secrets_for_screen_internal(screen_handles, from_active_screen, high16only, -1, secrets_do_replay_comment);
 	}
 	if (flags & mLIGHTBEAM)
 	{
 		for_every_rpos_in_screen(screen_handles, [&](const rpos_handle_t& rpos_handle) {
-			if (rpos_handle.layer != screen_handles[0].layer)
+			if (!do_layers && rpos_handle.layer != screen_handles[0].layer)
```
connorjclark, line 1282: drive-by fix...

src/dialog/combo_trigger_editor.cpp
```diff
@@ -156,7 +156,8 @@ static bool has_trigger_cause(combo_trigger const& trig)
 		TRIGFLAG_LWREFFIRE2,TRIGFLAG_SCREENLOAD,TRIGFLAG_SCREENUNLOAD,TRIGFLAG_PLAYERLANDHERE,
 		TRIGFLAG_PLAYERLANDANYWHERE,TRIGFLAG_CMBTYPECAUSES,
 		TRIGFLAG_BLOCK_TRIGGER_SAME_LAYER,TRIGFLAG_BLOCK_TRIGGER_ANY_LAYER,
-		TRIGFLAG_SWIMTRIG, TRIGFLAG_SWIMSENSTRIG,})) return true;
+		TRIGFLAG_SWIMTRIG, TRIGFLAG_SWIMSENSTRIG,
+		TRIGFLAG_MAPVIEW_ONLY, TRIGFLAG_MAPVIEW_NEVER,})) return true;
```
EmilyV99:
```suggestion
		TRIGFLAG_SWIMTRIG, TRIGFLAG_SWIMSENSTRIG,})) return true;
```
Conditions are not causes

src/dialog/combo_trigger_editor.cpp
```diff
@@ -616,6 +617,10 @@ std::shared_ptr<GUI::Widget> ComboTriggerDialog::view()
 								" after the warp effect; for scrolling, this is"
 								" before scrolling."),
 							TRIGFLAG(TRIGFLAG_SCREENUNLOAD, "Screen Unloads->"),
+							IBTN("Trigger condition: only occurs during Map View (spacebar map)."),
+							TRIGFLAG(TRIGFLAG_MAPVIEW_ONLY, "Map View Only"),
+							IBTN("Trigger condition: never occurs during Map View (spacebar map)."),
+							TRIGFLAG(TRIGFLAG_MAPVIEW_NEVER, "Map View Never"),
```
EmilyV99: Everything else in this section are Causes, should go somewhere else in the GUI (probably time for the `Conditions` tab to be split into sub-tabs; can take the current `Player Position` frame and make it it's own sub-tab, with the rest of the stuff there + these new ones forming an `Other` sub-tab)

src/zc/combos.cpp
```diff
@@ -1934,8 +1934,10 @@ static bool handle_trigger_conditionals(combined_handle_t const& comb_handle, si
 {
 	auto& cmb = comb_handle.combo();
 	if(cmb.triggers.size() <= idx) return false;
-	
+
 	auto& trig = cmb.triggers[idx];
+	if (g_is_map_view && trig.trigger_flags.get(TRIGFLAG_MAPVIEW_NEVER)) return false;
+	if (!g_is_map_view && trig.trigger_flags.get(TRIGFLAG_MAPVIEW_ONLY)) return false;
```
EmilyV99:
```suggestion
	if (trig.trigger_flags.get(g_is_map_view  ? TRIGFLAG_MAPVIEW_NEVER : TRIGFLAG_MAPVIEW_ONLY)) return false;
```
more compact

src/zc/maps.cpp
```diff
@@ -123,6 +128,15 @@ bool is_in_current_region(int map, int screen)
 
 bool is_in_current_region(int screen)
 {
+	// TODO: pretty hacky. Probably covers all cases of handling screen state / triggering for
+	// non-active screens, but can't be too certain. Map view obviously can show the same screen
+	// as an active screen (though it's a different mapscr object) - but I don't think anything else
+	// does in practice. MAPCOMBO3_impl usually is called on a neighboring, distinct screen; though
+	// with side warps that could happen to be the same screen as an active one.
```
EmilyV99, line 135: Even with side warps calling it on the same screen as an active one, the act of going through a warp means the version you're warping to is NOT the active version, so this seems fine.

src/zc/maps.cpp
```diff
@@ -6814,64 +6831,12 @@ std::array<mapscr, 7> loadscr2(int32_t screen)
...
-	}
+	g_is_map_view = true;
+	apply_state_changes_to_screen_handles(screen_handles, cur_map, screen, flags, true, true);
+	g_is_map_view = false;
 
-	clear_xdoors(screen_handles);
-	clear_xstatecombos(screen_handles);
-	
 	// check doors
 	if (isdungeon(screen))
```
EmilyV99, line 6841: Should dungeon doors also be handled as part of apply_state_changes_to_screen_handles? Unsure.

EmilyV99 (changes requested):

src/zc/maps.cpp
connorjclark: yes but the point is that this function merely checks the screen number - so for the case where you are on screen X and the side warp also goes to screen X: the state handling / triggering functions will incorrectly mark the side warp destination as being "active" and run code that should only run for the actually active region.

how this particular edge case is handled isn't new to this change, i'm just documenting it here. I fixed it for the map view case because that's where I saw a failure in a replay.

EmilyV99: I was agreeing with your handling of it lol 😅

EmilyV99 (approved):

PR #1372, opened 2026-06-12
Per combo pos states

EmilyV99:


src/zc/scripting/types/mapdata.cpp
```diff
@@ -1655,6 +1692,36 @@ static ffcdata* resolve_ffc_for_scripting_index(mapdata* mapdata, int index)
...
+		[](mapdata* mapdata, int index) -> int {
+			auto ffc_handle = resolve_ffc_handle_for_scripting_index(mapdata, index);
+			if (!ffc_handle)
+				return 0;
+			
+			int mi = mapind(ffc_handle.scr);
+			if (mi < 0)
+				return 0;
+
+			return game->ffcpos_states.get(mi).get(index % MAXFFCS);
```
connorjclark: index for ffcs here is 1-based - so need to be `(index - 1) % MAXFFCS`

src/zc/scripting/types/ffc.cpp
```diff
@@ -186,6 +186,28 @@ static ArrayRegistrar FFFLAGSD_registrar(FFFLAGSD, []{
 	return &impl;
 }());
 
+static ArrayRegistrar FFC_POSSTATE_registrar(FFC_POSSTATE, []{
+	static ScriptingArray_ObjectComputed<ffcdata, bool> impl(
+		[](ffcdata*){ return MAXFFCS; },
```
connorjclark: this should be _8_.

src/zc/scripting/types/ffc.cpp
```diff
@@ -186,6 +186,28 @@ static ArrayRegistrar FFFLAGSD_registrar(FFFLAGSD, []{
...
+		[](ffcdata* ffc, int index) -> bool {
+			int mi = mapind(cur_map, ffc->screen_spawned);
```
connorjclark: should home_screen be used here like elsewhere in the engine, depending on screen index?

src/zc/combos.cpp
```diff
@@ -2072,6 +2072,59 @@ bool force_ex_door_trigger_any(const combined_handle_t& handle, int dir, uint in
...
+{
+	auto& cmb = handle.combo();
+	if (cmb.triggers.size() <= idx) return false;
+	auto& trig = cmb.triggers[idx];
+	if (!trig.trigger_flags.get(TRIGFLAG_COMBOPOSSTATE_CAUSE))
+		return false;
+	
+	auto& pos_states = handle.is_ffc() ? game->ffcpos_states : game->pos_states;
+	auto mi = mapind(handle.base_scr());
+	auto posstate = pos_states[mi][handle.local_id()];
```
connorjclark: `.get()` b/c otherwise it will default insert won't it? and could add `get/set_combopos_state(handle, bit)` helper since this pattern is repeated a few times but slightly differently each time.

I think from this code (and maybe there's others like it?) alone, even if nothing uses pos states if a trigger happens we'll end up saving a bunch of useless data to the save file

src/zc/maps.cpp
```diff
@@ -1545,6 +1549,7 @@ static void apply_state_changes_to_screen_handles(const screen_handles_t& screen
 	int mi = mapind(map, screen);
 	clear_xdoors_mi(screen_handles, mi);
 	clear_xstatecombos_mi(screen_handles, mi);
+	clear_combopos_combos(screen_handles);
```
connorjclark: these methods need to pass-thru `mi`, otherwise when non-live code runs this function they use the wrong map screen index (whatever the current one is rather than `mi` here).

src/zc/hero.cpp
```diff
@@ -24141,6 +24141,7 @@ void HeroClass::checkspecial()
 
 		clear_xdoors(screen_handles, true);
 		clear_xstatecombos(screen_handles, true);
+		clear_combopos_combos(screen_handles);
```
connorjclark, line 24146: this runs for every rpos on every frame ...

guard on `game->pos_states.contains(mi) || ffcpos_states.contains(mi)`?

src/core/combo.cpp
```diff
@@ -994,6 +1011,21 @@ if (var > -2) \
...
+			if (do_set || do_unset)
+			{
+				string verb;
+				if (do_set && do_unset)
+					verb = "Toggle";
+				else if (do_set)
+					verb = "Set";
+				else
+					verb = "Unset";
+				effects << indent << fmt::format("{}s this combo position's PosState #{}", verb, int(combopos_state));
```
connorjclark: `\n` ?

src/zc/hero.cpp
connorjclark: i think i've misread this, this code is after enemies are defeated not every frame? maybe? or is it every frame if enemies are defeated? idk

either way, could clear_combopos_combos have an early exit if no pos state for this `mi`?

src/zc/maps.cpp
EmilyV99: Would they? I would assume the way I'm getting the `mi` from the `screen_handles` should prevent that...

src/zc/hero.cpp
EmilyV99: It is every frame, and a guard for that sounds smart!

src/zc/scripting/types/ffc.cpp
EmilyV99: OOPS.

src/zc/scripting/types/ffc.cpp
EmilyV99: Yes, probably. Maybe should be directly added to `mapind()`?
Doing that causes a bunch of failures and I don't feel like dealing with that rn, so just guarding the new code for now.

src/zc/maps.cpp
connorjclark: the problem is in `load_temp_mapscr_and_apply_secrets` when `layer >= 0` — at maps.cpp:1577, `source` gets rebound to the canonical screen the layer points at, and that copy becomes `screen_handles[0].base_scr`. canonical screens have their own map/screen fields, so recalculating mi from base_scr there gives the layer screen's mi - but live screens store pos states under the layer-0 screen's mi (rpos handles always use the layer-0 screen as base_scr). so they never match.

that's why `clear_xdoors_mi` / `clear_xstatecombos_mi` take mi as a param instead of recalculating - `apply_state_changes_to_screen_handles` already has the correct one at line 1549. `clear_combopos_combos` should do the same.

But...

there are other places in this PR where you calculate mi from the handle, and I guess threading `mi` to all of those would be too invasive...

maybe the fix is that `load_temp_mapscr_and_apply_secrets` after copying from `mapscr scr = *source` should set `scr->screen,map`? then could drop `mi` passed a param..

EmilyV99: ah. I don't think lying about screen,map is smart in the long run. just pushed passing `mi` through.

connorjclark: wait how'd that be lying? It's what is done when loading the temp screens for layers too iirc (loadscr)

EmilyV99: 👀 apparently it is . . . it seems weird to me that `->map`/`->screen` are for the base screen and not the mapscr in question

PR #1378, opened 2026-06-17
feat: larger than 1x1 liftable combos

EmilyV99:


src/zc/combos.cpp
```diff
@@ -1961,6 +2023,40 @@ static void trig_copycat(int cid, const combined_handle_t& handle, byte trigcopy
 	}
 }
 
+static map<std::pair<int,rpos_t>, vector<byte>> large_copycat_ids;
```
connorjclark, line 2026: please find an appropriate place to clear this - looks like entries are never removed (only removed from the inner vectors).

either on loadscr, or in init_game_vars.

if wanting to keep static could add a `combos_init_game_vars()` and call that in init_game_vars (like `maps_init_game_vars()` and `guys_init_game_vars()`)

src/core/qst_combos.cpp
```diff
@@ -912,6 +912,11 @@ int32_t readcombo_triggers_loop(PACKFILE* f, word s_version, combo_trigger& temp
 		if(!p_getc(&temp_trigger.combopos_state, f))
 			return 143;
 	}
+	if (s_version >= 71)
+	{
+		if(!p_getc(&temp_trigger.large_combo_copycat, f))
+			return 143;
```
connorjclark: nit: re-used code

src/dialog/comboeditor.cpp
```diff
@@ -3016,6 +3016,19 @@ std::shared_ptr<GUI::Widget> ComboEditorDialog::view()
 								{
 									SETFLAG(local_comboref.liftflags,LF_SPECIALITEM,state);
 								}
+							),
+							//
+							IBTN("All combos connected via 'Large Combo' settings on the General tab"
+								" will be lifted as one large object when any of them are lifted."
+								" Each combo will use it's own undercombo settings, but the top-leftmost"
```
connorjclark, line 3023: its

src/zc/combos.cpp
```diff
@@ -1898,6 +1898,68 @@ void trigger_save(newcombo const& cmb, mapscr* scr)
...
+	{
+		if (it->rpos == lowest_rpos)
+		{
+			if (it == handles.begin())
+				break; // already at front of list
+			// Move 'lowest' handle to front of list, so caller knows which handle
+			// to use for lifting flags.
+			rpos_handle_t lowest_handle = *it;
+			handles.erase(it);
+			handles.insert(handles.begin(), std::move(lowest_handle));
```
connorjclark, line 1957: doesn't matter but... there's nothing to move in rpos_handle_t :)

PR #1386, opened 2026-06-20
Weapon data scripting

EmilyV99:


src/zc/scripting/script_object.h
```diff
@@ -46,8 +46,8 @@ struct UserDataContainer
 
 	void clear()
 	{
-		auto ids = script_object_ids_by_type[type];
```
connorjclark, line 49: This is copied on purpose - delete_script_object can mutate the thing being iterated, hence the copy to prevent an error on mutation.

It's subtle, so please revert this change + add a comment saying this `auto ids` is copied on purpose b/c delete_script_object will mutate `script_object_ids_by_type`.

src/zc/ffscript.cpp
```diff
@@ -276,8 +294,8 @@ script_array* checkArray(uint32_t id, bool skipError)
 
 void script_bitmaps::update()
 {
-	auto ids = script_object_ids_by_type[user_bitmaps.type];
-	for (auto id : ids)
+	auto const& ids = script_object_ids_by_type[user_bitmaps.type];
```
connorjclark: same as above

src/core/combo.cpp
connorjclark, line 1: `newcombo::advpaste` should probably change to copy these shared_ptr objects into a new one, not share a reference.

as in `*lift_weap_data = *other.lift_weap_data` etc.

tests/replays/playground/auto_weapon_data_script_access.zplay
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
connorjclark: these "auto" replays are meant to to be generic scripts that are generated / managed by `tests/update_auto_script_tests.py` - seems like this one was manually made b/c I do not accompanying script file.

should be a generic script in the `tests/scripts/playground/auto/` folder + use test assertions to test specific behaviors. You just need to write the script file in that folder, then run the python script to generate the zplay for you.

EmilyV99: oh. well it required more setup than just a script, so sounds like an auto replay is not what I want then.

PR #1388, opened 2026-06-21
feat(zq): combo trigger copy/paste uses system clipboard + json

EmilyV99:


src/base/containers.cpp
```diff
@@ -1,5 +1,50 @@
...
+			else if (c >= 'A' && c <= 'F')
+				tmp = c - 'A';
+			else if (c >= 'a' && c <= 'f')
+				tmp = c - 'a';
+			else throw std::invalid_argument("bitstring must contain only valid hex characters!");
+			b |= tmp << offs;
+		}
+		inner_c.push_back(b);
+	}
+}
```
connorjclark, line 49: This code has a bug:

bitstring hex decoder drops the 0xA offset for letter digits, corrupting any nibble ≥ 10 on paste.

to_json encodes nibble values 10–15 as 'A'–'F' (char('A' + b - 0xA)). But from_json decodes them as:

```
else if (c >= 'A' && c <= 'F')
    tmp = c - 'A';        // 'A' -> 0, should be 0xA (10)
else if (c >= 'a' && c <= 'f')
    tmp = c - 'a';        // 'a' -> 0, should be 0xA (10)
```

Both branches are missing + 0xA. Failure scenario: copy a trigger whose req_global_state/req_screen_ex_door/etc. bitstring has any byte with a nibble ≥ 10 (e.g. a state byte of 0x0A, encoded as "A0"). On paste, 'A' decodes to 0, so the byte becomes 0x00 — the pasted trigger silently loses those state-requirement bits. Round-trip is only correct for bitstrings using digits 0–9. Fix: tmp = c - 'A' + 0xA; and tmp = c - 'a' + 0xA;.

Maybe add a test to src/test_runner/base_test_runner.cpp ?

connorjclark: Though honestly I'd have just encoded this is an integer array:

```
void to_json(json& j, const bitstring& b)   { j = b.cont.inner(); }
void from_json(const json& j, bitstring& b) { j.get_to(b.cont.mut_inner()); }
```

Avoids the hex parsing issues.

PR #1391, opened 2026-06-23
Script instance and export variables

EmilyV99:
Plans / Ideas:
1. Static Class Variables
2. Instance-based script variables
  - Initialize via metadata; use editor to edit values of specified variables per-object.

Implementation Details So Far:
- `static` can now be used on class variables, instead of only class functions.
- `static` / `nonstatic` can now be used on SCRIPT variables + functions.
  - By default, script members are static, unless the `#option DEFAULT_STATIC_SCRIPT_MEMBERS` is `off`.
  - non-static script members are only accessible by the script instance.
    This means non-static functions can't be called by other scripts,
    and non-static variables can't be read or changed by other scripts.
  - non-static script variables are per-script-instance. This means that
    ex. separate ffcs will have separate variables in their scripts.
  - All non-static variables/functions can be used from any non-static function in the script.
  - `this` is now considered a non-static script variable. This means it can be used
    from any non-static script function, instead of only in `void run()`.
  - non-static variables can have initializers, which must be compile-time constant, and are included in script metadata.
    - The plan is to allow overriding the initial values with in-editor values, but this requires a fair amount more work.

also, todo markers:
- debugger info isn't handling static functions very well (for scripts, and I would presume classes too)


src/components/zasm/eval.h
```diff
@@ -135,13 +135,15 @@ class VMInterface {
 
     virtual int32_t readStack(int32_t offset) = 0;
     virtual int32_t readGlobal(int32_t global_index) = 0;
+    virtual int32_t readScript(int32_t global_index) = 0;
```
connorjclark:
```suggestion
    virtual int32_t readScript(int32_t index) = 0;
```

src/parser/ByteCode.h
```diff
@@ -100,7 +100,7 @@ namespace ZScript {
 	typedef Opcode2<LSHIFTR32, VarArgument, VarArgument> O32BitLShiftRegister;
 	typedef Opcode2<LSHIFTV, VarArgument, LiteralArgument> OLShiftImmediate;
 	typedef Opcode2<LSHIFTV32, VarArgument, LiteralArgument> O32BitLShiftImmediate;
-	typedef Opcode2<MARK_TYPE_REG, GlobalArgument, TypeArgument> OMarkTypeRegister;
+	typedef Opcode2<MARK_TYPE_REG, VarArgument, TypeArgument> OMarkTypeRegister;
```
connorjclark, line 103: can delete these given your changes:

- GlobalArgument
- ScriptInstanceArgument
- caseGlobal/caseScriptInstance

tests/scripts/script_instance_vars/after.zs
```diff
@@ -0,0 +1,52 @@
+#option DEFAULT_STATIC_SCRIPT_MEMBERS off
```
connorjclark: These non-error scripts aren't great tests unless they actually run (compile tests isn't enough to be confident this new behavior is working). Can you move these to auto replay tests, and change the traces to asserts?

src/core/initdata.h
```diff
@@ -30,76 +30,80 @@ enum
...
 	zfix shove_offset = 6.5_zf;
 	zfix air_drag = 0.0100_zf;
 	
-	byte region_mapping;
+	byte region_mapping = REGION_MAPPING_FULL;
 	
 	word item_spawn_flicker = 32, item_timeout_dur = 512, item_timeout_flicker = 0;
 	byte item_flicker_speed = 2;
 	
 	int8_t hero_itembox_xofs, hero_itembox_yofs = 8;
```
connorjclark: missed one.

src/parser/ScriptParser.cpp
```diff
@@ -162,10 +162,21 @@ static unique_ptr<ScriptsData> _compile_helper(string const& filename, bool incl
 		}
 
 		FunctionData fd(program);
-		if(zscript_error_out) return result;
-		if (fd.globalVariables.size() > MAX_SCRIPT_REGISTERS)
+		if (auto count = fd.globalVariables.size(); count >= MAX_GLOBAL_VARIABLES)
```
connorjclark:
```suggestion
		if (auto count = fd.globalVariables.size(); count > MAX_GLOBAL_VARIABLES)
```

1024 globals is valid, yet this errors on the 1024th.

src/parser/ScriptParser.cpp
```diff
@@ -162,10 +162,21 @@ static unique_ptr<ScriptsData> _compile_helper(string const& filename, bool incl
...
+		{
+			log_error(CompileError::TooManyGlobal(NULL, count, MAX_GLOBAL_VARIABLES));
+			zscript_error_out = true;
+		}
+		for (Script* scr : program.scripts)
+		{
+			if (auto count = scr->instance_var_count(); count >= MAX_SCRIPT_INST_VARIABLES)
```
connorjclark:
```suggestion
			if (auto count = scr->instance_var_count(); count > MAX_SCRIPT_INST_VARIABLES)
```

same

src/parser/Scope.cpp
```diff
@@ -455,13 +461,34 @@ ZClass* ZScript::lookupClass(Scope const& scope, string const& name)
...
+			{
+				if (trimmed_match)
+					*trimmed_match = true;
+				temp = nullptr;
+			}
+		}
+		if (!temp && current->isClass() && trimClassLocal && trimmed_match)
+		{
+			auto* cscope = current->getClassScope();
+			*trimmed_match = cscope->getClassVar(name) != nullptr;
```
connorjclark: trimmed_match should probably only ever be set to true, never set unconditionally. eg `if (cscope->getClassVar(name)) *trimmed_match = true;`

The below example shows why.

```
class TestClass
{
	const int FOO = 5;
	static const int BAR = 6;
	int inst_var;

	void nonstatfn()
	{
		Trace(FOO);      // OK — plain const is fine from an instance method
		Trace(BAR);      // OK — static const is fine everywhere
		Trace(inst_var); // OK — instance member from instance method
	}
	static void statfn()
	{
		Trace(FOO);      // Error S009: Variable FOO has not been declared.
		                 //   ^ wrong twice: FOO *is* declared, and the intended
		                 //     message (S110) was not given
		Trace(BAR);      // OK — 'static const' compiles from a static method
		Trace(inst_var); // Error S110: Variable inst_var is not static, so cannot
		                 //   be accessed externally.  (correct message)
	}
}
```

Also, something to decide explicitly: should we just allow constants no matter their storage location from a static function? ie should `FOO` exist in `statfn`? After all, it being constant means it has _no real_ storage location, yet the choice of storage location determines whether it exists or not.

No strong opinion either way, but whatever you decide toss some of the above into one of your error test files.

docs-www/source/zscript/lang/options.rst
```diff
@@ -89,6 +90,14 @@ Standard Options
 	|                                    |                  |                       | warns/errors when a variable with a non-array type tries to     |
 	|                                    |                  |                       | hold an array.                                                  |
 	+------------------------------------+------------------+-----------------------+-----------------------------------------------------------------+
+	| DEFAULT_STATIC_SCRIPT_MEMBERS      | :zs_kw:`warn`    | |onoffauto|           | If :zs_kw:`auto`, picks between :zs_kw:`on` / :zs_kw:`off`      |
```
connorjclark: warn?

resources/include/bindings/itemdata.zh
```diff
@@ -451,11 +451,23 @@ class itemdata {
...
 
-	// The 8 InitD values shared by the item's Action, Pickup, and Sprite
-	// scripts.
+	// The 8 InitD values used by the Action script.
+	// [qr_SCRIPTS_SHARE_INITD] makes changing this array also change [PickupInitD] and [SpriteInitD].
 	//
 	// @zasm_var IDATAINITDD
 	internal untyped InitD[];
+	
+	// The 8 InitD values shared by the item's Pickup script.
```
connorjclark: shared -> used? and below

PR #1402, opened 2026-07-13
Scroll strings

EmilyV99:


src/core/msgstr.h
```diff
@@ -63,7 +88,7 @@ struct MsgStr
 	int16_t y;
 	uint16_t w;
 	uint16_t h;
-	word sfx;
+	word sfx, adv_sfx, menu_move_sfx, menu_close_sfx;
```
connorjclark, line 91: ensure all new fields are initialized in `clear()`.

src/gui/jwin.cpp
```diff
@@ -4827,22 +4829,15 @@ void _handle_jwin_scrollable_scroll_click(DIALOG *d, int32_t listsize, int32_t *
...
                 /* let other objects continue to animate */
-                int r = broadcast_dialog_message(MSG_IDLE, 0);
-                if (r & D_REDRAWME) should_redraw = true;
-
-                if (should_redraw)
-                {
-                    update_hw_screen();
-                }
+                broadcast_dialog_message(MSG_IDLE, 0);
+				update_hw_screen();
```
connorjclark, line 4840: flagging this as potentially performance regression (commit from 3 years ago - `fix: greatly reduce cpu usage when drawing dialogs`

https://github.com/ZQuestClassic/ZQuestClassic/commit/5d166d15f7d94c8bba4b28e3f8390b775d398284

what's the context for this change?

resources/include/bindings/common.zh
```diff
@@ -173,6 +173,24 @@ enum ShadowType
 	SHD_MAX
 };
 
+// Anchor positions relative to message boxes, used for icons
+// ([messagedata::MoreIconAnchor], [messagedata::ScrollUpIconAnchor], [messagedata::ScrollDownIconAnchor])
+// @versionadded 3.0
+enum MessageAnchor
+{
+	ANCH_TOP,
```
connorjclark: maybe `MSGANCH_*` or `MSG_ANCH_*`? Other messagedata enums I'm seeing all start with `MSG`

connorjclark: also, should move to messagedata.zh

src/gui/jwin.cpp
EmilyV99: draw changes occurring as a result of the MSG_IDLE via MSG_VSYNC weren't causing this to update_hw_screen, causing the scrolling preview to not work

This should only have any effect while the mouse is held down on a scrollbar

src/zq/gui/msgstr_preview.cpp
```diff
@@ -50,23 +205,61 @@ void put_msg_str([[maybe_unused]] char const* s, int32_t x, int32_t y, MsgStr co
...
+	auto bottom_margin_clip = [&]()
+	{
+		if (!get_qr(qr_STRINGS_DONT_SCROLL) && !get_qr(qr_OLD_STRING_EDITOR_MARGINS))
+			return false; // don't clip in scrolling mode
+		return !get_qr(qr_OLD_STRING_EDITOR_MARGINS)
+			&& cursor_y >= (h + (get_qr(qr_STRING_FRAME_OLD_WIDTH_HEIGHT)?16:0) - msg_margins[down]);
+	};
+	auto update_max_scroll = [&](word pos)
+	{
+		if (segmented_scroll && (pos % body_height))
```
connorjclark: can `body_height` be 0? potential div-by-zero error

src/zc/message_string.cpp
```diff
@@ -85,6 +96,81 @@ void draw_prt()
...
+	
+	destroy_bitmap(tmp_txt);
+	destroy_bitmap(tmp_menu);
+	
+	txt_bmp_height = new_height;
+}
+
+void update_max_scroll(word pos)
+{
+	if (segmented_scroll && (pos % body_height))
```
connorjclark: div-by-zero

src/zc/message_string.cpp
```diff
@@ -1437,23 +1567,70 @@ void tick_message()
...
+							target_scroll_pos += body_height;
+						else if (up)
+							target_scroll_pos -= body_height;
+					}
+					else
+					{
+						if (down)
+							scroll_pos += active_scroll_speed;
+						else if (up)
+							scroll_pos -= active_scroll_speed;
```
connorjclark: scroll_pos needs to be clamped to [0, max_scroll] after apply scroll speed - because the condition above is looking at `scroll_pos == target_scroll_pos` exactly. So if scroll speed does not exactly divide scroll height it won't end.

src/core/msgstr.h
```diff
@@ -43,6 +47,27 @@ struct message_portrait
...
+	screen,
+	// anchor x to screen + y to up
+	screen_y_offset,
+	
+	//
+	max_anchor
+};
+struct message_icon
+{
+	word sprite = 0;
```
connorjclark: I see we have compat code to set the "more" icon for existing strings; but I think this now means the default for new strings is to have no icon at positioned at 0, 0. is that what you intended? if not, should set the defaults for a new string ... somewhere (as initializers in the struct would be ok but somewhere else is probably better)

src/zq/gui/msgstr_preview.cpp
```diff
@@ -17,20 +17,175 @@ void init_msgstr(MsgStr *str);
...
+	can_scroll = !get_qr(qr_STRINGS_DONT_SCROLL) && !get_qr(qr_OLD_STRING_EDITOR_MARGINS);
+	setPreferredWidth(256_px*2+4_px + (can_scroll ? 18_px : 0_px));
+	setPreferredHeight(168_px*2+4_px);
+	bg_buf = create_bitmap_ex(8, 256, 168);
+	fg_buf = create_bitmap_ex(8, 256, 512);
+	clear_bitmap(bg_buf);
+	clear_bitmap(fg_buf);
+	fg_bmp_height = 512;
+	segmented_scroll = can_scroll && get_qr(qr_STRING_SEGMENTED_SCROLL);
+	memset(msg_margins, sizeof(msg_margins), 0);
```
connorjclark: wrong order for params

src/zc/scripting/types/messagedata.cpp
```diff
@@ -405,75 +595,265 @@ void messagedata_set_register(int32_t reg, int32_t value)
...
 			break;
 		}
+		case MESSAGEDATA_SHADOW_TYPE:
+		{
+			int32_t ID = GET_REF(msgdataref);
 
+			if(BC::checkMessage(ID) != SH::_NoError)
+				break;
+			else
+				MsgStrings[ID].shadow_type = ((byte)vbound((value/10000), 0, sstsMAX));
```
connorjclark:
```suggestion
				MsgStrings[ID].shadow_type = ((byte)vbound((value/10000), 0, sstsMAX-1));
```

src/zc/message_string.cpp
```diff
@@ -1693,21 +1958,57 @@ void dismiss()
...
+	}
 }
-//!TODO offer better user positioning options (ex. relative to bottom-left corner of string box)
-std::pair<int,int> more_xy()
+
+void draw_icons(bool drawPassiveSubscreenSeparate)
 {
-	int msgy = zc_min(160, (zinit.msg_more_is_offset==0)
-		? zinit.msg_more_y
-		: zinit.msg_more_y + get_str().y);
```
connorjclark, line 1707: is this clamp no longer needed? don't see anything like it in the new code

src/dialog/string_editor.cpp
```diff
@@ -44,13 +44,29 @@ GUI::ListData createShadowTypesListData()
...
+};
+
 StringEditorDialog::StringEditorDialog(size_t ind, int32_t templateID, int32_t addAfter)
 	: strIndex(ind), addAfter(addAfter),
 	tmpMsgStr(MsgStrings[ind]),
 	list_nextstr(GUI::ZCListData::strings(false)),
 	list_font(GUI::ZCListData::fonts(false,true,true)),
 	list_font_order(GUI::ZCListData::fonts(false,true,false)),
 	list_sfx(GUI::ZCListData::sfxnames(true)),
+	list_sprites(GUI::ZCListData::miscsprites()),
```
connorjclark: no icon is valid, right? do we need to present that as an option here?

the param for miscsprites is `skipNone = true` - but it uses -1 for "None" but you chose 0 to mean None here. so just setting to `false` here isn't the fix.

Does this suggest using 0 for "no icon" is the wrong approach?

src/zc/message_string.cpp
EmilyV99: I highly doubt it ever did anything in any actual quest, and barring someone actually presenting an issue with it, I think it's safe to just drop. It's arbitrarily assuming the full 16x16 of the sprite needs to be on-screen, which isn't necessarily the case if the actual non-transparent area of the sprite is smaller...

src/core/msgstr.h
EmilyV99: The default for a new string WOULD be that, and I don't see it as an issue. Most likely you'll have an existing string with an icon that you are using as a `Template:` string, in which case it would use that string's values as the new string defaults- if you aren't doing that, it can be assumed that you need to set up all the icons and such directly, and I see no reason to give it defaults based on old hardcoded values.

src/zc/message_string.cpp
EmilyV99: 🤔 I don't exactly follow the logic of `scroll_pos == target_scroll_pos` being an issue, as right below this segment it does `target_scroll_pos = scroll_pos;`. But, a clamp here seems correct regardless, so I see no reason to disagree with adding that.

src/dialog/string_editor.cpp
EmilyV99: Reasonable observation, perhaps `int16_t` instead of `word` would be more appropriate to allow `-1` for none

src/zc/message_string.cpp
connorjclark: oops, I cited the wrong condition. I meant to talk about the L1619 condition below.

The root of the problem is scroll_pos is unsigned, and we're not checking for underflow iiuc.

With speed 4 and max_scroll 10, scrolling up goes 10 → 6 → 2, and at 2 the `scroll_pos <= 0` check doesn't stop it so 2 - 4 wraps to 65534. Then down is blocked (>= max_scroll), A/B is blocked (cur_scroll() == max_scroll() at the advance check is false), and the box blits from outside the bitmap so it renders blank.

Would need to add a signed temp variable to hold the calculation.

```
int new_pos = int(scroll_pos) + (down ? int(active_scroll_speed) : -int(active_scroll_speed));
scroll_pos = vbound(new_pos, 0, int(max_scroll()));
target_scroll_pos = scroll_pos;
```

I believe the `segmented_scroll` branch is fine, but this last `else` branch needs this clamping.

EmilyV99: ookay this is just getting annoying, so I changed everything from fucking `word` to `int` lol. And I added a helper function for bounding the scroll.

connorjclark: good call

src/zq/gui/msgstr_preview.cpp
connorjclark: this doesn't look fixed - L208 can set body_height to 0

src/zq/gui/msgstr_preview.h
```diff
@@ -33,15 +28,28 @@ class MsgPreview: public Widget
...
 	}
 private:
-	std::string text;
 	DialogRef alDialog;
 	MsgStr const* str_data;
 	int32_t index;
+	BITMAP *bg_buf, *fg_buf;
+	
+	bool can_scroll, segmented_scroll;
+	word scroll_pos, body_height, max_visible_pos;
```
connorjclark: these are word, so that thing i pointed out in the player about underflow still happens in the preview ... can this code be de-duped?

src/zq/gui/msgstr_preview.cpp
```diff
@@ -17,20 +17,175 @@ void init_msgstr(MsgStr *str);
...
+				prv->is_tmp_scrolling = true;
+				prv->tmp_scroll = prv->scroll_pos;
+				word h = prv->body_height;
+				if (prv->segmented_scroll)
+				{
+					prv->tmp_scroll /= prv->body_height;
+					max /= prv->body_height;
+					h = 1;
+				}
+				if (_handle_jwin_scrollable_scroll_click(d, max+h, &prv->tmp_scroll, h))
```
connorjclark, line 110: doesn't this need to check `gui_mouse_x()` to scope to the scrollbar, rather than allow any click on the widget to scroll?

src/zq/gui/msgstr_preview.cpp
```diff
@@ -17,20 +17,175 @@ void init_msgstr(MsgStr *str);
...
+				prv->scroll_pos -= spd;
+				if (prv->scroll_pos < prv->target_scroll_pos)
+					prv->scroll_pos = prv->target_scroll_pos;
+			}
+			ret = D_REDRAWME;
+			break;
+		}
+		case MSG_DRAW:
+		{
+			prv->update_string();
```
connorjclark, line 67: this invalidates `max` captured in a local above.

src/zc/scripting/types/messagedata.cpp
```diff
@@ -405,75 +595,265 @@ void messagedata_set_register(int32_t reg, int32_t value)
...
+			break;
+		}
+		case MESSAGEDATA_ICON_MORE_SPRITE:
+		{
+			int32_t ID = GET_REF(msgdataref);
+
+			if(BC::checkMessage(ID) != SH::_NoError)
+				break;
+			else
+				MsgStrings[ID].icon_more.sprite = ((word)vbound((value/10000), 0, MAXSPRITES-1));
```
connorjclark: -1 is "none" now, right? need to allow in vbound

src/zc/message_string.cpp
```diff
@@ -1437,23 +1588,69 @@ void tick_message()
...
+							scroll_pos = scroll_bound(scroll_pos - active_scroll_speed);
+						target_scroll_pos = scroll_pos;
+					}
+				}
+			}
+			bool advance = do_end_str;
+			if (!advance && just_confirmed_menu)
 			{
+				target_scroll_pos = max_scroll();
+				advance = true;
```
connorjclark, line 1558: should `just_confirmed_menu` be cleared in here?

`\RunMenu \WaitAdvance \WaitAdvance` -> currently both wait advance would be skipped

EmilyV99: yep, nice catch!

src/zc/scripting/types/messagedata.cpp
EmilyV99: nice catch

src/zq/gui/msgstr_preview.h
EmilyV99: :sigh: not really in any easy way I can see

src/zc/message_string.cpp
EmilyV99: Moved this clamp to the quest load for the position which should catch older quests that bypassed this that had their more y set too high like the one you posted in discord.

connorjclark (approved):

PR #1404, opened 2026-07-15
fix(zc): big 4-directional enemy fire/attack frames overlapping walk tiles

connorjclark:
https://discord.com/channels/876899628556091432/1519766866027872347

### DIFF
--- a/src/zc/guys.cpp
+++ b/src/zc/guys.cpp
@@ -7400,17 +7400,19 @@ waves2:
 		tile+=2*f4;
 		if(clk2>0 && anim == a4FRM8DIRFB)                         //stopped to fire
 		{
-			tile+=80;
-			
+			// The 8 walking directions of a big enemy span 8 tile rows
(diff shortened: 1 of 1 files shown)


src/zc/guys.cpp
```diff
@@ -7405,17 +7405,20 @@ void enemy::update_enemy_frame()
...
+			int32_t fire_off = get_qr(qr_BROKEN_BIG_ENEMY_FIRE_ANIM) ? 80 : 160;
+			tile+=fire_off;
+
 			if(clk2<17)                                           //firing
 			{
-				tile+=80;
+				tile+=fire_off;
 			}
 		}
 		ignore_extend = true;
```
EmilyV99, line 7412:
```suggestion
			tile+=80;

			if(clk2<17)                                           //firing
			{
				tile+=80;
			}
		}
		ignore_extend = get_qr(qr_BROKEN_BIG_ENEMY_FIRE_ANIM);
```
Does this just work, I wonder?

PR #1407, opened 2026-07-16
feat(zscript): @ExportInitD0 - @ExportInitD7

EmilyV99:
Using multiple annotations for one thing is a bit bulky. With this new annotation, specify all the information about an InitD at once! ex. `@ExportInitD0("Speed", "Movement Speed in px/frame")` to specify a name + helptext. A 3rd parameter can be supplied to specify the field type. This combines `@InitD0()`, `@InitDHelp0()`, and `@InitDType0()` into a single annotation.

end changelog

This refactors the annotation code to handle different numbers of parameters, instead of being limited to only exactly 0 or 1 parameter.

### DIFF
--- a/src/parser/ffscript.ypp
+++ b/src/parser/ffscript.ypp
@@ -18,8 +18,9 @@
 #include "zsyssimple.h"
 #include "parser/ParserHelper.h"
 #include "base/util.h"
+#include "base/headers.h"
+#include <fmt/ranges.h>
 
-using std::string;
 using std::ostringstream;
 using namespace ZScript;
 
@@ -52,7 +53,7 @@ void trunc_str(std::string& str, size_t sz, std::string const& header, int32_t r
 	}
 }
 
-enum
+enum AnnotationParamType
 {
 	ANNTY_NONE,
 	ANNTY_STR,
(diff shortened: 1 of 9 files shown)


src/parser/ffscript.ypp
```diff
@@ -1209,26 +1325,68 @@ Annotated_Script :
...
+							}
+							
+							if (sz >= 2)
+							{
+								auto& param = data.params[1];
+								annot_trunc_str(param.val, 65535);
+								script->metadata.initd_help[c] = param.val;
+							}
+							
+							// if (sz >= 1)
```
connorjclark: if you want to keep the block indentation for style purposes, might as well uncomment even if the conditional is always true. Looks less like a mistake.

src/parser/ffscript.ypp
```diff
@@ -64,91 +65,223 @@ static string annot_tys[ANNTY_MAX] = {"Empty", "String", "Number"};
...
+
+static optional<int8_t> annot_param_check_swaptype(AnnotData& data, size_t param_idx)
 {
-	if(expected == data.type)
-		return true;
-	annot_errstr("ERROR: Bad Annotation Value: @"+data.key
-		+" expects a "+annot_tys[expected]+", not a "+annot_tys[data.type], data.location);
-	return false;
+	auto& pdata = data.params[param_idx];
+	assert(pdata.type == ANNTY_STR);
```
connorjclark: please use CHECK instead of assert here and elsewhere - better error message.

connorjclark (approved):
Can you add a simple script `tests/scripts/misc/annotations.zs` to use this new annotation? Can also test the new error cases you added in `tests/scripts/errors/errors_annotations.zs`.

connorjclark (approved):

PR #1411, opened 2026-07-27
qr: Disable f6 + docs stuff

EmilyV99:


src/dialog/info.cpp
```diff
@@ -48,59 +47,10 @@ void InfoDialog::postinit()
...
-					break;
-				case INFOHINT_T_RULETMPL:
-					ruleTemplates.insert(val);
-					break;
-			}
-		}
-		#endif
-	}
+	
+	auto hints = parse_hint_text(d_text);
```
connorjclark: is it ok to be calling this in zplayer? before it was editor only

scripts/generate_zscript_api_docs.py
```diff
@@ -308,9 +310,10 @@ def format_comment(text: str) -> str:
 
     if has_comment_or_interesting_tags:
         add('')
-        add('.. rst-class:: classref-comment')
+        add('.. container:: classref-comment')
         add('')
-        for tag, value in symbol.comment.tags:
+        # Sort 'deprecated' tags to the end
+        for tag, value in sorted(symbol.comment.tags, key=lambda s: ('deprecated' in s, s)):
```
connorjclark:
```suggestion
        for tag, value in sorted(symbol.comment.tags, key=lambda s: 'deprecated' in s[0]):
```

`s` is a tuple - so this was checking if "deprecated" exactly existed in the tuple. But that misses soft_deprecated, etc.

also removed the `, s` part to prefer authored ordering (python sorted is a stable sort) rather than arbitrary alpha-ordering

EmilyV99: nice catch, and as for ordering I thought a more stable order would be better- but will defer to you here

PR #1414, opened 2026-08-03
Improve Exports

EmilyV99:


docs-www/source/zscript/lang/scripts.rst
connorjclark, line 45: we should add a note that this is 3.0 only (true for most of the content on this page).

src/base/general.h
connorjclark:
```suggestion
	static constexpr byte DISABLED = 0x1;
```

fixes build

src/dialog/script_data_editor.cpp
```diff
@@ -174,42 +199,380 @@ std::shared_ptr<GUI::Widget> ScriptDataDialog::view()
...
+								string key;
+								if (decimal)
+									key = fmt::format("{} ({})", str, val);
+								else
+									key = fmt::format("{} ({:0{}})", str, val.getInt(), digits);
+								vals[key] = val;
+							}
+							
+							for(auto& [name, val] : vals)
+								ld.add(name, val);
```
connorjclark: does this work? current code truncates to just the integer part.

i think:

```suggestion
								ld.add(name, val.getZLong());
```

src/dialog/script_data_editor.cpp
```diff
@@ -174,42 +199,380 @@ std::shared_ptr<GUI::Widget> ScriptDataDialog::view()
...
+							}
+							if (clist.empty())
+								clist.emplace_back(CheckListInfo::DISABLED, "No Flags Found");
+							
+							auto btn = Button(text = "P",
+								hAlign = 0.0, padding = 0_px,
+								width = 24_px, height = 24_px,
+								onPressFunc = [&, idx, set_arg]()
+								{
+									int flags = args[idx];
```
connorjclark, line 549: how does this handle when the bitflags are long vs int?

don't we sometimes need a `/ 10000`?

resources/include/bindings/common.zh
```diff
@@ -1046,3 +1046,15 @@ enum CornerIndex
 	CORNER_BL,
 	CORNER_BR,
 };
+
+@Bitflags("long")
+enum EnumExportDisplayFlags
```
connorjclark, line 1052: @versionadded 3.0

src/parser/ZScript.cpp
```diff
@@ -260,16 +260,84 @@ void UserScript::register_instance_var(Variable* v, optional<int32_t> value)
...
+	// are always in the same file as others in the same instance.
+	if (loc1.first_line < loc2.first_line)
+		return true;
+	if (loc2.first_line < loc1.first_line)
+		return false;
+	if (loc1.first_column < loc2.first_column)
+		return true;
+	if (loc2.first_column < loc1.first_column)
+		return false;
+	return true;
```
connorjclark: Returning true when the items are equal is UB for `std::sort` (it expects a strict weak ordering).

```suggestion
	return false;
```

or replace most of this with `return std::tie(...) < std::tie(...)` (fill in the things you are comparing.

src/parser/RegBaseVisitor.cpp
```diff
@@ -1021,6 +1130,408 @@ bool RegBaseVisitor::parse_annotations_data(ASTDataDeclList& node)
...
+			zfix list_key = zslongToFix(*val);
+			if (bit_mode != ASTDataEnum::BIT_NONE)
+			{
+				if (_check_power_2_value(list_key, bit_mode == ASTDataEnum::BIT_LONG))
+				{
+					// not a power of 2, cannot display
+					continue;
+				}
+			}
+			if (node.export_data.custom_export_names.contains(list_key))
```
connorjclark:
```suggestion
			if (name_map.contains(list_key))
```

src/parser/RegBaseVisitor.cpp
```diff
@@ -1021,6 +1130,408 @@ bool RegBaseVisitor::parse_annotations_data(ASTDataDeclList& node)
...
+			assert(decl && decl->manager);
+			Constant const* c = dynamic_cast<Constant const*>(decl->manager);
+			assert(c);
+			auto val = c->getCompileTimeValue();
+			auto name = c->getName();
+			assert(val && name);
+			
+			// this check is explicitly BEFORE some options are trimmed; the prefix should account
+			// for all constants in the enum, not only the displayable ones.
+			if (prefix)
```
connorjclark: for `{FOO_BAR, FOO_BAZ}` this would find `FOO_B`, not what we want.

correct code for this is in `ASTDataEnum::getDocumentationPrefix`.

src/parser/RegBaseVisitor.cpp
```diff
@@ -1021,6 +1130,408 @@ bool RegBaseVisitor::parse_annotations_data(ASTDataDeclList& node)
...
+						if (node.export_data.custom_export_names.contains(list_key))
+						{
+							handleError(CompileError::AnnotationError(&annot,
+								fmt::format("Annotation '@{}' found multiple strings for value {}!", key, list_key)));
+							failed = true;
+							break;
+						}
+						node.export_data.custom_export_names[list_key] = param.str;
+						bool overflow = list_key >= 0b100000000000000000;
+						list_key <<= 1;
```
connorjclark: this would overflow on the 18th flag

src/parser/RegBaseVisitor.cpp
```diff
@@ -1021,6 +1130,408 @@ bool RegBaseVisitor::parse_annotations_data(ASTDataDeclList& node)
...
+				idx = (special_engine_export)q;
+				break;
+			}
+		}
+		if (idx != special_engine_export::none)
+			node.export_data.engine_type = idx;
+		else
+		{
+			handleError(CompileError::AnnotationError(&node,
+				fmt::format("Annotation '@{}' got \"{}\", but expected one of [{}]", key, param.str, fmt::join(special_engine_export_names, ", ")).c_str()));
```
connorjclark: this includes "NONE" - is that intended?

src/dialog/script_data_editor.cpp
```diff
@@ -174,42 +199,380 @@ std::shared_ptr<GUI::Widget> ScriptDataDialog::view()
...
+				}
+				else if (expdata.export_custom_type != var_custom_export_type::none)
+				{
+					bool is_long = false;
+					switch (expdata.export_custom_type)
+					{
+						case var_custom_export_type::custom_dropdown:
+						{
+							auto& ld = widget_data.list;
+							map<string, zfix> vals;
```
connorjclark, line 467: this loses authored order and forces alpha order - use `std::unordered_map` instead?

docs-www/source/zscript/lang/scripts.rst
```diff
@@ -48,33 +48,96 @@ also allows "resetting to default" an exported variable's value.
...
+	|                       | 'String' gives a name to the current 'value', and                     | Ex. `@ExportDropdown("Apple", "Banana", 6, "Grape")`  |
+	|                       | then increments the 'value' by 1.                                     | would specify "Apple" as 0, "Banana" as 1, and        |
+	|                       |                                                                       | "Grape" as 6.                                         |
+	+-----------------------+-----------------------------------------------------------------------+-------------------------------------------------------+
+	| `@ExportBitflags`     | Takes many parameters.                                                | Sets the exported value to use flag checkboxes,       |
+	|                       | Specifying no parameters will automatically name                      | displaying a checkbox for each specified option.      |
+	|                       | every possible flag by it's number.                                   |                                                       |
```
connorjclark: it's -> its.

here and below.

connorjclark (commented):
I pushed a commit to add some tests here. They add coverage for many of the comments I left.

src/base/general.h
EmilyV99: I knew it was gonna be something like this but nothing I tried worked lol, didn't try `constexpr`

src/dialog/script_data_editor.cpp
EmilyV99: No, the `args` value here is always `*10000`. The long values are handling their offset above, on the line that reads `zfix v = is_long ? zslongToFix(1 << q) : zfix(1 << q);`.

src/parser/RegBaseVisitor.cpp
EmilyV99: After the 18th flag, it does overflow, yes. It won't print an error if that's the last flag though, because `(q + 1) < num_params` is a condition on the overflow error.

src/parser/RegBaseVisitor.cpp
EmilyV99: uhggggg, no. that's annoying.

src/dialog/script_data_editor.cpp
EmilyV99: Dropdowns generally want to be alphabetized for the type-to-search to function properly... so unsure which is more important to preserve here.

src/parser/RegBaseVisitor.cpp
connorjclark: Overflow is UB, so it has to be guarded, not just detected for the error message. I added a fixup commit to address.

tests/scripts/errors/errors_exports.zs
```diff
@@ -0,0 +1,99 @@
...
+	nonstatic int btn_conflict;
+
+	// Engine value is incompatible with bool
+	@Export("Bool Engine"),
+	@ExportEngineValue("Color")
+	nonstatic bool bool_engine;
+
+	// Param must name an enum type
+	@Export("Not An Enum"),
+	@ExportEnum(NOT_AN_ENUM)
```
connorjclark: `@ExportEnum(NOT_AN_ENUM)` does not produce an error.

src/dialog/script_data_editor.cpp
connorjclark: I'm not sure exactly how the code is wrong but I think it's how `flags` is used.

Given this code:

```c
@Export("Flags"),
@ExportBitflags("Up", "Down", 16, "Big")
nonstatic int flag_var = 3;   // Up|Down
```

The default-enabled checkboxes in the editor are just "Big" - rather than Up and Down.

fix may be something like this?

```
int flags = is_long ? args[idx] : args[idx] / 10000;
...
set_arg(is_long ? flags : flags * 10000);
```

connorjclark (commented):
`882a6aeac6 build!: make checklist constant constexpr` needs to be re-ordered to prevent a broken bisect. Can either move it _before_ `feat(zscript)!:  @ExportDropdown` or squash into that commit.

tests/scripts/errors/errors_exports.zs
EmilyV99: this is EXCEEDINGLY confusing because I get an error for it in my local test

connorjclark: It doesn't error in tests/scripts/errors/errors_exports.zs (so it's not doing anything there as a regression check) - but it does error if I move it to a script with otherwise no errors. so it's a problem with an earlier compiler pass ending the error checking before it gets to this.

EmilyV99: Oh, wait- some of your errors there are registration-time errors, but it can't know at registration-time that the identifier is an invalid type. (A type and a variable/constant with the same name can co-exist, so your constant does not rule out that there's also an enum of that name which has simply yet-to-be-registered).

So, aye, no bug here, we just need a second error file to handle SemanticAnalyzer-time errors which has no RegistrationVisitor-time errors in it.

EDIT: Like what you just pushed, exactly lol.

src/dialog/script_data_editor.cpp
EmilyV99: yep, got it. Thanks 😅

connorjclark (approved):

PR #1415, opened 2026-08-03
fix(zc)!: enemy 'Fire Sound' now plays for every weapon type

connorjclark:


connorjclark:
https://github.com/ZQuestClassic/ZQuestClassic/pull/1419

PR #1417, opened 2026-08-06
feat(zscript): Annotations can use constants

EmilyV99:


src/parser/RegBaseVisitor.cpp
```diff
@@ -419,4 +413,892 @@ void RegBaseVisitor::initInternalVar(ASTDataDeclList* node)
...
+	if (breakRecursion()) return false;
+	if (!node.script_annotation) return true;
+	
+	static vector<string> valid_keys = {
+		"Author",
+		"InitScript",
+		"ScriptInfo",
+		"ScriptSetup",
+	};
+	static bool initialized = false;
```
connorjclark, line 1037: this is never set to true anywhere

EmilyV99: OH! That'd be a problem lmao! nice catch
Wouldn't break logically because it's just a contains check so extra entries wouldn't break, but, would waste loads of memory ballooning that vector 😅

src/parser/AST.h
```diff
@@ -1081,6 +1114,8 @@ namespace ZScript
 		BitMode getBitMode() const {return bitmode;}
 		
 		optional<zfix> increment_val;
+		
+		owning_ptr<ASTAnnotationList> enum_annotation {};
```
connorjclark, line 1118: ASTDataDeclList and ASTDataEnum aren't copying data_annotation/enum_annotation in their custom copy ctors.

connorjclark (commented):
Seems named enums have a problem.

```
const int FOO = 10;
@Increment(FOO)
enum blah
{
	A, B, C, D
};
CONST_ASSERT(A == 0 && B == 10 && C == 20 && D == 30); // fails assert
```

---

Order seems to effect how values are resolved in annotations. Add a debug print to `UserScript::register_instance_var` and you'll see the following doesn't resolve the expected range:

```
generic script exportAfter {
	@Export("Value"),
	@ExportRange(MIN, MAX)
	nonstatic int val = 1;

	const int MIN = -5;   // declared after the var
	const int MAX = 5;
}
```

---

`@AlwaysRunEndpoint("garbage") loop() {}` doesn't emit any error, even though it's invalid.

